PR #7, opened 2024-08-16
Test requirements

chrismeyersfsu:
* Add requirements file for testing. Maybe the same requirements file should be added for packaging?
* Thanks for the "hello world" tests. I think I kept the spirit of them but modified them to invoke the newly added modules
* Deleted "hello world" module example.


tests/importable_test.py
```diff
@@ -19,8 +20,31 @@ def test_entry_points_exposed(entry_points_group: str) -> None:
     pre-installed.
     """
     entry_points = _discover_entry_points(group=entry_points_group)
-    assert 'x' in entry_points.names
+    expected_entry_points = {
+        'aim': 'aim_plugin',
```
webknjaz, line 24: @chrismeyersfsu FYI such things should go into either parametrize or `pytest-subtest` so that they show up separately in the test report.

tests/importable_test.py
```diff
@@ -19,8 +20,31 @@ def test_entry_points_exposed(entry_points_group: str) -> None:
...
+        'hashivault_kv': 'hashivault_kv_plugin',
+        'hashivault_ssh': 'hashivault_ssh_plugin',
+        'azure_kv': 'azure_keyvault_plugin',
+        'centrify_vault_kv': 'centrify_plugin',
+        'thycotic_dsv': 'dsv_plugin',
+        'thycotic_tss': 'tss_plugin',
+        'aws_secretsmanager_credential': 'aws_secretmanager_plugin',
+    }
+    for x in expected_entry_points.keys():
+        assert x in entry_points.names
```
webknjaz, line 35: if you were to avoid subtests, though, the correct check would be a oneliner `assert set(expected_entry_points.keys()) <= set(entry_points.names)`

PR #19, opened 2024-08-27
Initial pre-commit conformity

chrismeyersfsu:
I ran `tox -e pre-commit` locally and these are the changes it made. Once this lands, other PR's shouldn't get the additional pre-commit github job generated commit.


src/awx_plugins/credentials/azure_kv.py
```diff
@@ -1,63 +1,81 @@
...
-    'required': ['url', 'client', 'secret', 'tenant', 'secret_field'],
-}
+clouds = [
+    vars(azure_cloud)[n] for n in dir(azure_cloud)
+    if n.startswith('AZURE_') and n.endswith('_CLOUD')
+]
+default_cloud = vars(azure_cloud)['AZURE_PUBLIC_CLOUD']
+
+
+azure_keyvault_inputs = {'fields': [{'id': 'url',
```
webknjaz, line 16: Weird that it made this edit. It's probably because of mixed style inside (perhaps those inline dicts). It should be possible to dedent it if you want.

webknjaz: Looks like long unsplittable lines were also causing this.

I've made https://github.com/ansible/awx-plugins/pull/20 to show how the diff may be smaller.

Also, this PR still triggers formatting (https://results.pre-commit.ci/run/github/836873755/1724796257.Vt0Lwn2FSQ-J0oVorFr3bw) while mine does not — it needs a couple of reruns to settle, between several tools changing files.

webknjaz:
I'll go with #20 instead, as it's cleaner and does address those problems with double formatting conflicts.

PR #22, opened 2024-08-28
Show pre-commit bot vs command line conflict

chrismeyersfsu:
Running `tox -e pre-commit` does _not_ reformat the code. Let's see what the bot does.

### DIFF
--- a/src/awx_plugins/credentials/aim.py
+++ b/src/awx_plugins/credentials/aim.py
@@ -91,7 +91,8 @@ def aim_backend(**kwargs):
         webservice_id = 'AIMWebService'
 
     requests.get(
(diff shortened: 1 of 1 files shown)


chrismeyersfsu:
Hmm, seems I was wrong. Closing this.

webknjaz:
@chrismeyersfsu there was a case when two runs of pre-commit were needed to settle the formatting between several tools. In that case, the pre-commit.ci bot would send the first formatting update but not the following one, not reacting to its own commit (GitHub Apps normally do this to prevent infinite loops where they trigger themselves).

I handled this manually once, which is why you're no longer seeing it.

PR #24, opened 2024-08-29
Add assumerole

thedoubl3j:
Adding in the AWS assume role cred plugin authored by @derekwaters found in this AWX [PR](https://github.com/ansible/awx/pull/15294).


pyproject.toml
```diff
@@ -71,6 +71,7 @@ x = "awx_plugins.credentials.x.api:XPlugin"
 [project.entry-points."awx.credential_plugins"]  # pre-existing entry points group name
 x = "awx_plugins.credentials.x.api:XPlugin"
 # conjur = awx.main.credential_plugins.conjur:conjur_plugin
+aws_assumerole = awx_plugins.credentials.aws_assumerole:aws_assumerole_plugin
```
webknjaz: [Syntax error](https://app.readthedocs.org/projects/awx-plugins-core/builds/25474010/)
```suggestion
aws_assumerole = "awx_plugins.credentials.aws_assumerole:aws_assumerole_plugin"
```

tests/credential_plugins_test.py
```diff
@@ -131,6 +132,67 @@ def test_hashivault_handle_auth_not_enough_args():
...
 
 
+def test_aws_assumerole_with_accesssecret():
+    kwargs = {
+        'access_key': 'my_access_key',
+        'secret_key': 'my_secret_key',
+        'role_arn': 'the_arn',
+        'identifier': 'access_token',
+    }
+    with mock.patch.object(aws_assumerole, 'aws_assumerole_getcreds') as method_mock:
```
webknjaz: In `pytest,` it's customary to use `monkeypatch`. If `mock` is needed, it should be integrated through the `pytest-mock` plugin and injected as a `mocker` fixture.

tests/credential_plugins_test.py
```diff
@@ -131,6 +132,67 @@ def test_hashivault_handle_auth_not_enough_args():
         hashivault.handle_auth()
 
 
+def test_aws_assumerole_with_accesssecret():
```
webknjaz: To avoid adding more linting failures, this should have a good docstring.

pyproject.toml
webknjaz, line 1: I see undeclared dependencies — they also need to go here.

webknjaz:
@thedoubl3j It should be possible to rebase now. Check what jobs are green on `devel` and make sure they stay that way in the PR. Currently, that's all the docs jobs. The tests pass, but the jobs are red because of low code coverage.

pyproject.toml
```diff
@@ -71,6 +71,7 @@ x = "awx_plugins.credentials.x.api:XPlugin"
 [project.entry-points."awx.credential_plugins"]  # pre-existing entry points group name
 x = "awx_plugins.credentials.x.api:XPlugin"
 # conjur = awx.main.credential_plugins.conjur:conjur_plugin
+"aws_assumerole = awx_plugins.credentials.aws_assumerole:aws_assumerole_plugin"
```
webknjaz: Still a syntax error. The previous suggestion was correct:

```suggestion
aws_assumerole = "awx_plugins.credentials.aws_assumerole:aws_assumerole_plugin"
```

Just click the `Commit suggestion` button on GH UI next time. It'll make a proper commit, additionally preserving the credit.

thedoubl3j: ack

webknjaz:
FYI, PEP 257-compliant docstrings are always made of triple double-quotes.

pyproject.toml
webknjaz: @thedoubl3j did you accidentally remove the entry point?

thedoubl3j: @webknjaz yup, started rebasing and getting pulled multiple dirs. patch inc

.flake8
```diff
@@ -89,6 +89,11 @@ per-file-ignores =
   # have <= members in tests (WPS202), including many local vars (WPS210),
   # additionally test docstrings don't need param lists (DAR, DCO020):
   tests/**.py: DAR, DCO020, S101, S404, S603, WPS202, WPS210, WPS436, WPS442, WPS450
+  # (B950) is ignored due to it erroring on help text which is displayed in the API.
+  # While they can be shorter, they are not a priority to shorten.
+  #(WPS226) is ignored as it is erroring on the inputs field.
```
webknjaz:
```suggestion

  # (WPS226) is ignored as it is erroring on the inputs field.
```

.flake8
```diff
@@ -89,6 +89,11 @@ per-file-ignores =
   # have <= members in tests (WPS202), including many local vars (WPS210),
   # additionally test docstrings don't need param lists (DAR, DCO020):
   tests/**.py: DAR, DCO020, S101, S404, S603, WPS202, WPS210, WPS436, WPS442, WPS450
```
webknjaz: let's add a separator line.
```suggestion
  tests/**.py: DAR, DCO020, S101, S404, S603, WPS202, WPS210, WPS436, WPS442, WPS450

```

.flake8
```diff
@@ -89,6 +89,11 @@ per-file-ignores =
...
+  # (B950) is ignored due to it erroring on help text which is displayed in the API.
+  # While they can be shorter, they are not a priority to shorten.
```
webknjaz:
```suggestion
```

This rule specifically already allows going 10% over the line length limit in a few places. But not in most of a module's contents.

I think I mentioned this elsewhere already — I expect that newly added code would not increase the number of linting violations.

I've fixed this in some (most?) places already. This is not because of the help text, this is because of the heavy indentation chosen. If you move the braces on separate lines, I'm pretty sure `autopep8` will reformat it for you, similar to https://github.com/ansible/awx-plugins/blob/82c8c95/src/awx_plugins/credentials/dsv.py#L8-L82. I noticed that it suggests highly indented formatting if you don't do this.
Hinting `autopep8` what the formatting preference are like that, can help you save some manual formatting efforts if it goes into wrong direction.

In general, long lines suggest that code might be getting hard to consume, so I recommend avoiding ignoring such rules at all cost.

.flake8
```diff
@@ -89,6 +89,11 @@ per-file-ignores =
...
+  #(WPS226) is ignored as it is erroring on the inputs field.
+  # This current pattern is used across all plugins, we might consider extending this ignore to all plugins.
```
webknjaz: It'd be also useful to add context on what the rule demands since, unlike pylint, flake8 does not name alphanumeric identifiers and it's usually hard to guess what's ignored by staring at the numbers.

tests/credential_plugins_test.py
```diff
@@ -131,6 +132,76 @@ def test_hashivault_handle_auth_not_enough_args():
         hashivault.handle_auth()
 
 
+def test_aws_assumerole_with_accesssecret(monkeypatch):
```
webknjaz: While on it, please add typing
```suggestion
def test_aws_assumerole_with_accesssecret(monkeypatch: pytest.MonkeyPatch) -> None:
```

tests/credential_plugins_test.py
```diff
@@ -131,6 +132,76 @@ def test_hashivault_handle_auth_not_enough_args():
...
+    kwargs['identifier'] = 'secret_key'
+    token = aws_assumerole.aws_assumerole_backend(**kwargs)
+    assert token == 'the_secret_key'
+
+    kwargs['identifier'] = 'access_key'
+    token = aws_assumerole.aws_assumerole_backend(**kwargs)
+    assert token == 'the_access_key'
+
+
+def test_aws_assumerole_with_arnonly(monkeypatch):
```
webknjaz:
```suggestion
def test_aws_assumerole_with_arnonly(monkeypatch: pytest.MonkeyPatch) -> None:
```

pyproject.toml
```diff
@@ -27,6 +27,7 @@ dependencies = [  # runtime deps  # https://packaging.python.org/en/latest/guide
   "python-dsv-sdk >= 1.0.4",  # credentials.thycotic_dsv
   "python-tss-sdk >= 1.2.1",  # credentials.thycotic_tss
   "requests",  # credentials.aim, credentials.centrify_vault, credentials.conjur, credentials.hashivault
+  "datetime", # credentials.aws_assume_role
```
webknjaz: 1) the comments have 2 spaces before `#`, follow the style around
2) I was going to suggest keeping the deps sorted (it looks like only pyyaml is out of order for some reason, but...
3) I don't think this is the right (or needed for that matter) dependency — https://pypi.org/p/DateTime is a Zope project. And looking at the actual usage, I'm pretty sure you're only importing the stdlib `datetime` module: I haven't seen any APIs that aren't present @ https://docs.python.org/3.11/library/datetime.html.

```suggestion
```

src/awx_plugins/credentials/aws_assumerole.py
```diff
@@ -0,0 +1,128 @@
...
+                                    'label': 'Identifier',
+                                    'type': 'string',
+                                    'help_text': _('The name of the key in the assumed AWS role to fetch [AccessKeyId | SecretAccessKey | SessionToken].'),
+                                    },
+                                   ],
+                      'required': ['role_arn'],
+                      }
+
+
+def aws_assumerole_getcreds(access_key, secret_key, role_arn, external_id):
```
webknjaz: plz add typing to all new code. If I were to guess, this is probably something like
```suggestion
def aws_assumerole_getcreds(access_key: str | None, secret_key: str | None, role_arn: str | None, external_id: int) -> dict:
```

.flake8
```diff
@@ -89,6 +89,11 @@ per-file-ignores =
...
+  src/awx_plugins/aws_assumerole.py: B950, WPS226
```
webknjaz:
```suggestion
  src/awx_plugins/aws_assumerole.py: WPS226
```

src/awx_plugins/credentials/aws_assumerole.py
```diff
@@ -0,0 +1,128 @@
...
+        )
+    except ClientError as ce:
+        raise ValueError(f'Got a bad client response from AWS: {ce.msg}.')
+
+    credentials = response.get('Credentials', {})
+
+    return credentials
+
+
+def aws_assumerole_backend(**kwargs):
```
webknjaz: Plz type this too: https://typing.rtfd.io/en/latest/spec/callables.html#unpack-kwargs / https://docs.python.org/3.11/library/typing.html#typing.Unpack.
*(can it have regular explicit args listed instead?)*

src/awx_plugins/credentials/aws_assumerole.py
```diff
@@ -0,0 +1,128 @@
...
+    ):
+        # Connect using credentials in the EE
+        connection = boto3.client(service_name='sts')
+    else:
+        # Connect to AWS using provided credentials
+        connection = boto3.client(
+            service_name='sts',
+            aws_access_key_id=access_key,
+            aws_secret_access_key=secret_key,
+        )
```
webknjaz: Would this be simpler?
```suggestion
    explicit_credentials_empty = not access_key and not secret_key
    credential_kwargs = {} if explicit_credentials_empty else {
        # EE creds are read from the env
        'aws_access_key_id': access_key,
        'aws_secret_access_key': secret_key,
    }
    connection = boto3.client(service_name='sts', **credential_kwargs)
```

src/awx_plugins/credentials/aws_assumerole.py
```diff
@@ -0,0 +1,128 @@
...
+    credentials = _aws_cred_cache.get(credential_key, None)
+
+    # If there are no credentials for this user/ARN *or* the credentials
+    # we have in the cache have expired, then we need to contact AWS again.
+    #
+    if (credentials is None) or (
+        credentials['Expiration'] < datetime.datetime.now(
+            credentials['Expiration'].tzinfo,
+        )
+    ):
```
webknjaz: Readability tip: labeling parts of conditionals with named variables eliminates the need for using code comments for story-telling.
```suggestion
    creds_expired = (
        (creds_expire_at := credentials.get('Expiration')) and
        creds_expire_at < datetime.now(credentials['Expiration'].tzinfo)
    )
    if creds_expired:
```
(this suggestion requires `from datetime import datetime` at the top, which would also simplify a number of places by dropping those confusing `datetime.datetime.` references)

src/awx_plugins/credentials/aws_assumerole.py
```diff
@@ -0,0 +1,128 @@
...
+        )
+
+        _aws_cred_cache[credential_key] = credentials
+
+    credentials = _aws_cred_cache.get(credential_key, None)
+
+    if identifier in credentials:
+        return credentials[identifier]
+
+    raise ValueError(f'Could not find a value for {identifier}.')
```
webknjaz: This could be a nice way to chain the root cause in tracebacks:
```suggestion
    try:
        return credentials[identifier]
    except KeyError as key_err:
        raise ValueError(f'Could not find a value for {identifier}.') from key_err
```

tests/credential_plugins_test.py
```diff
@@ -131,6 +132,76 @@ def test_hashivault_handle_auth_not_enough_args():
...
+    token = aws_assumerole.aws_assumerole_backend(**kwargs)
+    assert token == 'the_access_token'
+
+    kwargs['identifier'] = 'secret_key'
+    token = aws_assumerole.aws_assumerole_backend(**kwargs)
+    assert token == 'the_secret_key'
+
+    kwargs['identifier'] = 'access_key'
+    token = aws_assumerole.aws_assumerole_backend(**kwargs)
+    assert token == 'the_access_key'
```
webknjaz: These are 3 separate tests that should show up separately in the pytest's test report.
This could be implemented by `pytest-subtest` but a better way to do this is using `@pytest.mark.parametrize`. Please move these into params.

tests/credential_plugins_test.py
```diff
@@ -131,6 +132,76 @@ def test_hashivault_handle_auth_not_enough_args():
...
```
webknjaz: Same here: these are 3 tests, not one. Split them accordingly.

webknjaz:
@thedoubl3j plz also add the entry point to the smoke test params @ https://github.com/ansible/awx-plugins/blob/82c8c95/tests/importable_test.py#L27-L73

.flake8
thedoubl3j: change committed per previous review.

.flake8
thedoubl3j: added in next commit.

.flake8
webknjaz: Looks like it was never actually pushed?

webknjaz:
Rebase this. From now on, all the GHA checks are expected to show up green.

src/awx_plugins/credentials/aws_assumerole.py
thedoubl3j: will be in the next commit

src/awx_plugins/credentials/aws_assumerole.py
thedoubl3j: you are correct, bit more easy to read and simpler

src/awx_plugins/credentials/aws_assumerole.py
thedoubl3j: can you explain this a bit further? my evaluation of this is that cred_expire_at is being declared before use. try to hack together something simple in a python environment and I got a NameError
```pycon
>>> b = (c:= 1 and c)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
NameError: name 'c' is not defined
```

I think I understand where you are coming from a readability standpoint vs code comments.

src/awx_plugins/credentials/aws_assumerole.py
thedoubl3j: interesting, have never seen `from key_err` before. TIL

tests/credential_plugins_test.py
thedoubl3j: fixed and combined with previous test in next commit

tests/credential_plugins_test.py
thedoubl3j: done

src/awx_plugins/credentials/aws_assumerole.py
```diff
@@ -0,0 +1,143 @@
...
+
+from django.utils.translation import gettext_lazy as _
+
+import boto3
+
+from .plugin import CredentialPlugin
+
+
+try:
+    from botocore.exceptions import ClientError
```
webknjaz: Can we make this a dependency so that it's always importable and wouldn't need to be suppressed?

src/awx_plugins/credentials/aws_assumerole.py
```diff
@@ -0,0 +1,143 @@
+import datetime
+import hashlib
+
+from django.utils.translation import gettext_lazy as _
```
webknjaz: FTR, a parallel effort moves such things into the interfaces repo temporarily. Although, the thing I was migrating was `gettext_noop`. I wonder if only one of the function variants needs to be used (noop or lazy)...

Anyway, at some point this will probably have to become
```suggestion
from awx_plugins.interfaces._temporary_private_api import gettext_noop
```
or
```suggestion
from awx_plugins.interfaces._temporary_private_api import gettext_lazy
```

src/awx_plugins/credentials/aws_assumerole.py
webknjaz: @thedoubl3j I misplaced the parens in the code suggestion, it seems. The assignment expression needs to be wrapped:
```pycon
$ python3.13t
Python 3.13.0rc1 experimental free-threading build (main, Aug 13 2024, 04:25:17) [GCC 13.3.1 20240614] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> b = (c:= 1 and c)
Traceback (most recent call last):
  File "<python-input-4>", line 1, in <module>
    b = (c:= 1 and c)
                   ^
NameError: name 'c' is not defined
>>> b = (c:= 1) and c
>>> b
1
>>> c
1
```

I've updated the above suggested comment, and it should be ready to be accepted by clicking a button on the GH UI.

src/awx_plugins/credentials/aws_assumerole.py
webknjaz: Yeah, it was introduced around Python 3.3, I think. It sets an explicit `__cause__` attribute, which is then used instead of the implicitly set `__context__`. Many people consider this the best practice because you're vocal about the causes of exceptions.
The traceback shows both exceptions, but the text is slightly different. If you use `from`, it'll say that the previous exception in the trace was the direct cause. You can also use `from None` to discard the exception from the previous context so that it's not attached to the reraised one.

src/awx_plugins/credentials/aws_assumerole.py
thedoubl3j: I think we could. My guess is that this was probably in place given how EEs function in AWX but given the context of how we bring in plugins like this, it would have been added to the AWX so this would have been removed at some point if we were not splitting out modules. 

Will add the dep and remove the warning. This also looks like a pattern that might been copied from the AWS collection since I believe they handle their dep requirements in catches like this.

src/awx_plugins/credentials/aws_assumerole.py
```diff
@@ -0,0 +1,143 @@
...
+        role_arn: str | None,
+        external_id: int,
+) -> dict:
+    explicit_credentials_empty = not access_key and not secret_key
+    credential_kwargs = {} if explicit_credentials_empty else {
+        # EE creds are read from the env
+        'aws_access_key_id': access_key,
+        'aws_secret_access_key': secret_key,
+    }
+    connection = boto3.client(service_name='sts', **credential_kwargs)
```
webknjaz: > [!TIP]
> Try changing https://github.com/ansible/awx-plugins/blob/072b41b/.pre-commit-config.yaml#L238C7-L238C18 to `boto3-stubs[sts]` (in all three MyPy invocations in that config).

src/awx_plugins/credentials/aws_assumerole.py
thedoubl3j: ahh, that is what I wasn't seeing. cheers good sir. added.

src/awx_plugins/credentials/aws_assumerole.py
thedoubl3j: currently I think I have from `.plugin import CredentialPlugin, translate_function as _`. I will rewrite to point to the temp interface.

webknjaz: The temporary private module has been renamed. Wait for a few hours to see the finalized version revealed.

.flake8
```diff
@@ -96,6 +96,7 @@ per-file-ignores =
   # The following ignores must be fixed and the entries removed from this config:
   src/awx_plugins/credentials/aim.py: ANN003, ANN201, B950, CCR001, D100, D103, LN001, Q003, WPS210, WPS221, WPS223, WPS226, WPS231, WPS336, WPS432
   src/awx_plugins/credentials/aws_secretsmanager.py: ANN003, ANN201, D100, D103, WPS111, WPS210, WPS226, WPS329, WPS529
+  src/awx_plugins/credentials/aws_assumerole.py: WPS226, WPS332
```
webknjaz, line 116: I've disabled WPS332 globally and tweaked WPS226 to have a higher threshold. So this entry is no longer necessary.
```suggestion
```

tests/importable_test.py
```diff
@@ -70,6 +70,12 @@ def __str__(self) -> str:
         'aws_secretsmanager_credential',
         'awx_plugins.credentials.aws_secretsmanager:aws_secretmanager_plugin',
     ),
+    EntryPointParam(
+        'awx_plugins.credentials',
+        'aws_assumerole',
+        'awx_plugins.credentials:aws_assumerole_plugin',
+    ),
+
```
webknjaz:
```suggestion
```

tests/credential_plugins_test.py
```diff
@@ -134,6 +135,75 @@ def test_hashivault_handle_auth_not_enough_args() -> None:
...
+            },
+        ),
+        (
+            {
+                'role_arn': 'the_arn',
+                'identifier': 'access_token',
+            },
+        ),
+    ),
+)
```
webknjaz: This seems too nested:
```suggestion
    'kwargs',
    (
        {
            'access_key': 'my_access_key',
            'secret_key': 'my_secret_key',
            'role_arn': 'the_arn',
            'identifier': 'access_token',
        },
    ),
    (
        {
            'role_arn': 'the_arn',
            'identifier': 'access_token',
        },
    ),
)
```
Also, perhaps rename the argument to be more explicit about what it is?

tests/credential_plugins_test.py
```diff
@@ -134,6 +135,75 @@ def test_hashivault_handle_auth_not_enough_args() -> None:
...
+        'the_access_token',
+    ),
+    (
+        'access_key',
+        'the_access_key',
+    ),
+    (
+        'secret_key',
+        'the_secret_key',
+    ),
```
webknjaz: And this one is not nested enough (causing that cryptic exception in pytest and pylint invocations). Let's see if this helps:
```suggestion
    (
	    (None, 'the_access_token'),
	    ('access_key', 'the_access_key'),
	    ('secret_key', 'the_secret_key'),
    ),
```

pyproject.toml
```diff
@@ -27,6 +27,7 @@ dependencies = [  # runtime deps  # https://packaging.python.org/en/latest/guide
   "python-dsv-sdk >= 1.0.4",  # credentials.thycotic_dsv
   "python-tss-sdk >= 1.2.1",  # credentials.thycotic_tss
   "requests",  # credentials.aim, credentials.centrify_vault, credentials.conjur, credentials.hashivault
+  "botocore",  # credentials.aws_assume_role
```
webknjaz:
```suggestion
  "botocore",  # credentials.aws_assume_role, credentials.awx_secretsmanager
```

webknjaz: Could you also add `credentials.aws_assume_role` to the comment next to `boto3` above?

src/awx_plugins/credentials/aws_assumerole.py
```diff
@@ -0,0 +1,165 @@
...
+    :rtype: dict
+    :raises ValueError: If the client response is bad.
+    """
+    explicit_credentials_empty = not access_key and not secret_key
+    credential_kwargs = {} if explicit_credentials_empty else {
+        # EE creds are read from the env
+        'aws_access_key_id': access_key,
+        'aws_secret_access_key': secret_key,
+    }
+    connection = boto3.client(service_name='sts', **credential_kwargs)
```
webknjaz: So I looked into https://github.com/boto/boto3/blob/07050a2/boto3/session.py#L51-L52 and these args already default to `None`. Thus, I believe that the special-casing is not actually necessary on our side in this case:
```suggestion
    from mypy_boto3_sts.client import STSClient

    connection: STSClient = boto3.client(
        service_name='sts',
        # The following EE creds are read from the env if they are not passed:
        aws_access_key_id=access_key,  # defaults to `None` in the lib
        aws_secret_access_key=secret_key,  # defaults to `None` in the lib
    )
```

*(the only difference is that the `explicit_credentials_empty` check was sensitive to other falsy values like an empty string)*

The type import can be moved to the top after accepting this suggestion. It's probably good to put it under `if TYPE_CHECKING` so there's no runtime dependency, only typecheck-time. https://youtype.github.io/boto3_stubs_docs/mypy_boto3_sts/client/#stsclient_1.

src/awx_plugins/credentials/aws_assumerole.py
```diff
@@ -0,0 +1,165 @@
...
+    try:
+        response = connection.assume_role(
+            RoleArn=role_arn,
+            RoleSessionName='AAP_AWS_Role_Session1',
+            ExternalId=external_id,
+        )
+    except ClientError as ce:
+        raise ValueError(f'Got a bad client response from AWS: {ce.message}.')
```
webknjaz: It's usually better to chain the exceptions explicitly and chose more verbose var names:
```suggestion
    except ClientError as client_err:
        raise ValueError(
            f'Got a bad client response from AWS: {ce.message}.',
        ) from client_err
```

src/awx_plugins/credentials/aws_assumerole.py
```diff
@@ -0,0 +1,165 @@
...
+    """
+    explicit_credentials_empty = not access_key and not secret_key
+    credential_kwargs = {} if explicit_credentials_empty else {
+        # EE creds are read from the env
+        'aws_access_key_id': access_key,
+        'aws_secret_access_key': secret_key,
+    }
+    connection = boto3.client(service_name='sts', **credential_kwargs)
+    try:
+        response = connection.assume_role(
```
webknjaz:
```suggestion
        from mypy_boto3_sts.type_defs import AssumeRoleResponseTypeDef

        response: AssumeRoleResponseTypeDef = connection.assume_role(
```
https://youtype.github.io/boto3_stubs_docs/mypy_boto3_sts/client/#assume_role
https://youtype.github.io/boto3_stubs_docs/mypy_boto3_sts/type_defs/#assumeroleresponsetypedef

src/awx_plugins/credentials/aws_assumerole.py
webknjaz: Though, this https://youtype.github.io/boto3_stubs_docs/mypy_boto3_sts/usage/#implicit-type-annotations says that the type defs can be implicit.

.flake8
```diff
@@ -111,20 +111,22 @@ per-file-ignores =
...
+  src/awx_plugins/credentials/injectors.py: ANN001, ANN201, ANN202, C408, D100, D103, WPS110, WPS111, WPS202, WPS210, WPS226, WPS347, WPS433, WPS440
   src/awx_plugins/credentials/plugin.py: ANN001, ANN002, ANN101, ANN201, ANN204, B010, D100, D101, D103, D105, D107, D205, D400, E731, WPS115, WPS432, WPS433, WPS440, WPS442, WPS601
   src/awx_plugins/credentials/plugins.py: B950,D100, D101, D103, D105, D107, D205, D400, LN001, WPS204, WPS229, WPS433, WPS440
   src/awx_plugins/credentials/tss.py: ANN003, ANN201, D100, D103, E712, WPS433, WPS440, WPS503
   src/awx_plugins/inventory/plugins.py: ANN001, ANN002, ANN003, ANN101, ANN102, ANN201, ANN202, ANN206, B950, C812, C819, D100, D101, D102, D205, D209, D400, D401, LN001, LN002, N801, WPS110, WPS111, WPS202, WPS210, WPS214, WPS301, WPS319, WPS324, WPS331, WPS336, WPS337, WPS338, WPS347, WPS421, WPS433, WPS450, WPS510, WPS529
   tests/credential_plugins_test.py: ANN101, B017, C419, D100, D102, D103, D205, D209, D400, DAR, PT011, S105, WPS111, WPS117, WPS118, WPS202, WPS352, WPS421, WPS433, WPS507
   tests/importable_test.py: ANN101, DAR
+  tests/credential_plugins_test.py: DAR, WPS430
```
webknjaz: I'm pretty sure two lines won't work for the same file. I don't think they'd be merged properly. I'm not sure which one ends up being selected, but line 127 already has these two codes suppressed, so duplicating it wouldn't be necessary anyway:
```suggestion
```

pyproject.toml
```diff
@@ -27,6 +27,7 @@ dependencies = [  # runtime deps  # https://packaging.python.org/en/latest/guide
   "python-dsv-sdk >= 1.0.4",  # credentials.thycotic_dsv
   "python-tss-sdk >= 1.2.1",  # credentials.thycotic_tss
   "requests",  # credentials.aim, credentials.centrify_vault, credentials.conjur, credentials.hashivault
+  "botocore",  # credentials.aws_assume_role, credentials.awx_secretsmanager
```
webknjaz: Could you extend the comment next to `boto3` as well?

src/awx_plugins/credentials/aws_assumerole.py
```diff
@@ -0,0 +1,170 @@
...
+    ],
+}
+
+
+def aws_assumerole_getcreds(
+        access_key: str | None,
+        secret_key: str | None,
+        role_arn: str | None,
+        external_id: int,
+) -> dict:
```
webknjaz:
```suggestion
) -> CredentialsTypeDef | dict[Never, Never]:
```
(import the types at the top, perhaps under type-checking only, since the first one is coming from the typing-only dep that's not in the runtime requirements; import annotations from future to make this work w/o having to quote the annotation, alternatively, wrap it with quotes)

src/awx_plugins/credentials/aws_assumerole.py
```diff
@@ -0,0 +1,170 @@
...
+    from mypy_boto3_sts.client import STSClient
+
+    connection: STSClient = boto3.client(
+        service_name='sts',
+        # The following EE creds are read from the env if they are not passed:
+        aws_access_key_id=access_key,  # defaults to `None` in the lib
+        aws_secret_access_key=secret_key,  # defaults to `None` in the lib
+    )
+    try:
+        from mypy_boto3_sts.type_defs import AssumeRoleResponseTypeDef
```
webknjaz: This will need to be moved to the top of the module. It'll also generate an import error if not made conditional (via `typing.TYPE_CHECKING`)

src/awx_plugins/credentials/aws_assumerole.py
```diff
@@ -0,0 +1,170 @@
...
+    )
+    if creds_expired:
+
+        credentials = aws_assumerole_getcreds(
+            access_key, secret_key, role_arn, external_id,
+        )
+
+        _aws_cred_cache[credential_key] = credentials
+
+    credentials = _aws_cred_cache.get(credential_key, None)
```
webknjaz: This shouldn't use `None` as a fallback. There's a bug here. If the lookup does indeed return `None`, the item access below would be `None[identifier]` which won't generate a `KeyError` but a `TypeError`. So it won't be caught and converted into a `ValueError` but instead bubble up the stack as a `TypeError` and will be messy to process on the calling side.
Use a thing that does indeed have a `__getitem__()` to prevent that:
```suggestion
    credentials = _aws_cred_cache.get(credential_key, {})
```
(or you could even move that into the return line with a double `[]`, and any of these would cause a `KeyError` that would be processed the same way, unless somebody manages to put an illegal type into the cache, of course)

src/awx_plugins/credentials/aws_assumerole.py
```diff
@@ -0,0 +1,170 @@
...
+import datetime
+import hashlib
+
+import boto3
+from botocore.exceptions import ClientError
+
+from .plugin import CredentialPlugin, translate_function as _
+
+
+_aws_cred_cache = {}
```
webknjaz: This seems to be
```suggestion
_aws_cred_cache: dict[str, CredentialsTypeDef | dict[Never, Never]] | dict[Never, Never] = {}
```

(provided you import the types, I think this should work)

src/awx_plugins/credentials/aws_assumerole.py
```diff
@@ -0,0 +1,170 @@
+"""This module provides integration with AWS AssumeRole functionality."""
+
+import datetime
+import hashlib
+
+import boto3
+from botocore.exceptions import ClientError
+
+from .plugin import CredentialPlugin, translate_function as _
```
webknjaz: This is the new import:
```suggestion
from awx_plugins.interfaces._temporary_private_django_api import (  # noqa: WPS436
    gettext_noop as _,
)

from .plugin import CredentialPlugin
```

This should fix the import error here: https://github.com/ansible/awx-plugins/actions/runs/11226550925/job/31207297417?pr=24#step:18:62.

src/awx_plugins/credentials/aws_assumerole.py
```diff
@@ -0,0 +1,170 @@
...
+        from mypy_boto3_sts.type_defs import AssumeRoleResponseTypeDef
+
+        response: AssumeRoleResponseTypeDef = connection.assume_role(
+            RoleArn=role_arn,
+            RoleSessionName='AAP_AWS_Role_Session1',
+            ExternalId=external_id,
+        )
+    except ClientError as client_err:
+        raise ValueError(
+            f'Got a bad client response from AWS: {ce.message}.',
```
webknjaz: The linter caught this: https://github.com/ansible/awx-plugins/actions/runs/11226550925/job/31207297582?pr=24#step:18:83
```suggestion
            f'Got a bad client response from AWS: {client_err.message}.',
```

.flake8
webknjaz: Judging by https://github.com/ansible/awx-plugins/actions/runs/11226550925/job/31207297582?pr=24#step:18:88 reporting `WPS430`, I'd say that the first exact path match is being used, and this one isn't.
It has `DAR` in that list but not `WPS430`.

.flake8
```diff
@@ -111,20 +111,22 @@ per-file-ignores =
...
+  src/awx_plugins/credentials/injectors.py: ANN001, ANN201, ANN202, C408, D100, D103, WPS110, WPS111, WPS202, WPS210, WPS226, WPS347, WPS433, WPS440
   src/awx_plugins/credentials/plugin.py: ANN001, ANN002, ANN101, ANN201, ANN204, B010, D100, D101, D103, D105, D107, D205, D400, E731, WPS115, WPS432, WPS433, WPS440, WPS442, WPS601
   src/awx_plugins/credentials/plugins.py: B950,D100, D101, D103, D105, D107, D205, D400, LN001, WPS204, WPS229, WPS433, WPS440
   src/awx_plugins/credentials/tss.py: ANN003, ANN201, D100, D103, E712, WPS433, WPS440, WPS503
   src/awx_plugins/inventory/plugins.py: ANN001, ANN002, ANN003, ANN101, ANN102, ANN201, ANN202, ANN206, B950, C812, C819, D100, D101, D102, D205, D209, D400, D401, LN001, LN002, N801, WPS110, WPS111, WPS202, WPS210, WPS214, WPS301, WPS319, WPS324, WPS331, WPS336, WPS337, WPS338, WPS347, WPS421, WPS433, WPS450, WPS510, WPS529
   tests/credential_plugins_test.py: ANN101, B017, C419, D100, D102, D103, D205, D209, D400, DAR, PT011, S105, WPS111, WPS117, WPS118, WPS202, WPS352, WPS421, WPS433, WPS507
```
webknjaz:
```suggestion
  tests/credential_plugins_test.py: ANN101, B017, C419, D100, D102, D103, D205, D209, D400, DAR, PT011, S105, WPS111, WPS117, WPS118, WPS202, WPS352, WPS421, WPS430, WPS433, WPS507
```

tests/credential_plugins_test.py
```diff
@@ -134,6 +135,67 @@ def test_hashivault_handle_auth_not_enough_args() -> None:
...
+    (
+        (None, 'the_access_token'),
+        ('access_key', 'the_access_key'),
+        ('secret_key', 'the_secret_key'),
+    ),
+)
+def test_aws_assumerole_identifier(
+    monkeypatch: pytest.MonkeyPatch,
+    kwargs: dict, identifier_key: str, expected: dict[str, str],
```
webknjaz: Looking into the decorators above, the actual types are
```suggestion
    kwargs: dict[str, str], identifier_key: str | None, expected: str,
```

tests/credential_plugins_test.py
```diff
@@ -134,6 +135,67 @@ def test_hashivault_handle_auth_not_enough_args() -> None:
...
+    monkeypatch.setattr(
+        aws_assumerole,
+        'aws_assumerole_getcreds',
+        mock_getcreds,
+    )
+
+    if identifier_key:
+        kwargs['identifier'] = identifier_key
+
+    token = aws_assumerole.aws_assumerole_backend(**kwargs)
```
webknjaz: I think it's cleaner to have an additional var here instead of mutating the inputs:
```suggestion
    extra_kwargs = {'identifier': identifier_key} if identifier_key else {}

    token = aws_assumerole.aws_assumerole_backend(**kwargs, **extra_kwargs)
```

tests/credential_plugins_test.py
```diff
@@ -134,6 +135,67 @@ def test_hashivault_handle_auth_not_enough_args() -> None:
...
+    'kwargs',
+    (
+        {
+            'access_key': 'my_access_key',
+            'secret_key': 'my_secret_key',
+            'role_arn': 'the_arn',
+            'identifier': 'access_token',
+        },
+    ),
+    (
```
webknjaz:
```suggestion
```

src/awx_plugins/credentials/aws_assumerole.py
thedoubl3j: done

src/awx_plugins/credentials/aws_assumerole.py
```diff
@@ -0,0 +1,177 @@
...
+
+if typing.TYPE_CHECKING:
+    from mypy_boto3_sts.client import STSClient
+    from mypy_boto3_sts.type_defs import AssumeRoleResponseTypeDef
+
+from .plugin import CredentialPlugin
+
+
+_aws_cred_cache: dict[str, CredentialsTypeDef |
+                      dict[Never, Never]] | dict[Never, Never] = {}
```
webknjaz:
```suggestion
_aws_cred_cache: dict[
    str,
    CredentialsTypeDef | dict[typing.Never, typing.Never],
] | dict[typing.Never, typing.Never] = {}
```

src/awx_plugins/credentials/aws_assumerole.py
```diff
@@ -0,0 +1,177 @@
...
+    ],
+}
+
+
+def aws_assumerole_getcreds(
+        access_key: str | None,
+        secret_key: str | None,
+        role_arn: str | None,
+        external_id: int,
+) -> CredentialsTypeDef | dict[Never, Never]:
```
webknjaz:
```suggestion
) -> CredentialsTypeDef | dict[typing.Never, typing.Never]:
```

src/awx_plugins/credentials/aws_assumerole.py
```diff
@@ -0,0 +1,179 @@
...
+        aws_secret_access_key=secret_key,  # defaults to `None` in the lib
+    )
+    try:
+        response: AssumeRoleResponseTypeDef = connection.assume_role(
+            RoleArn=role_arn,
+            RoleSessionName='AAP_AWS_Role_Session1',
+            ExternalId=external_id,
+        )
+    except ClientError as client_err:
+        raise ValueError(
```
webknjaz, line 113: `ValueError` does not seem like a good fit for this kind of error. It's probably a `RuntimeError` or something like that.
```suggestion
        raise RuntimeError(
```
Or perhaps a
```suggestion
        raise LookupError(
```
It might also be a
```suggestion
        raise ConnectionError(
```

`ValueError` would signal to the caller that they passed something wrong into the function call. But this message seems to imply it's an unexpected platform problem or something like that.

src/awx_plugins/credentials/aws_assumerole.py
```diff
@@ -0,0 +1,179 @@
...
+        )
+
+        _aws_cred_cache[credential_key] = credentials
+
+    credentials = _aws_cred_cache.get(credential_key, {})
+
+    try:
+        return credentials[identifier]
+    except KeyError as key_err:
+        raise ValueError(
```
webknjaz, line 173: The same concern goes for this exception. Although I imagine that awx expects a `ValueError` so we can't change it w/o defining the interface first...

src/awx_plugins/credentials/aws_assumerole.py
```diff
@@ -0,0 +1,179 @@
...
+    gettext_noop as _,
+)
+
+import boto3
+from botocore.exceptions import ClientError
+
+
+if typing.TYPE_CHECKING:
+    from mypy_boto3_sts.client import STSClient
+    from mypy_boto3_sts.type_defs import AssumeRoleResponseTypeDef
```
webknjaz:
```suggestion
    from mypy_boto3_sts.type_defs import AssumeRoleResponseTypeDef, CredentialsTypeDef
```

src/awx_plugins/credentials/aws_assumerole.py
```diff
@@ -0,0 +1,179 @@
...
+    gettext_noop as _,
+)
+
+import boto3
+from botocore.exceptions import ClientError
+
+
+if typing.TYPE_CHECKING:
+    from mypy_boto3_sts.client import STSClient
+    from mypy_boto3_sts.type_defs import AssumeRoleResponseTypeDef, CredentialsTypeDef
```
webknjaz:
```suggestion
    from mypy_boto3_sts.type_defs import (
        AssumeRoleResponseTypeDef,
        CredentialsTypeDef,
    )
```

src/awx_plugins/credentials/aws_assumerole.py
```diff
@@ -0,0 +1,182 @@
...
+    ],
+}
+
+
+def aws_assumerole_getcreds(
+        access_key: str | None,
+        secret_key: str | None,
+        role_arn: str | None,
+        external_id: int,
+) -> CredentialsTypeDef | dict[typing.Never, typing.Never]:
```
webknjaz:
```suggestion
) -> "CredentialsTypeDef" | dict[typing.Never, typing.Never]:
```

src/awx_plugins/credentials/aws_assumerole.py
```diff
@@ -0,0 +1,182 @@
...
+        AssumeRoleResponseTypeDef,
+        CredentialsTypeDef,
+    )
+
+from .plugin import CredentialPlugin
+
+
+_aws_cred_cache: dict[
+    str,
+    CredentialsTypeDef | dict[typing.Never, typing.Never],
```
webknjaz:
```suggestion
    "CredentialsTypeDef" | dict[typing.Never, typing.Never],
```

src/awx_plugins/credentials/aws_assumerole.py
```diff
@@ -0,0 +1,182 @@
...
+    :raises ValueError: If the client response is bad.
+    """
+    connection: STSClient = boto3.client(
+        service_name='sts',
+        # The following EE creds are read from the env if they are not passed:
+        aws_access_key_id=access_key,  # defaults to `None` in the lib
+        aws_secret_access_key=secret_key,  # defaults to `None` in the lib
+    )
+    try:
+        response: AssumeRoleResponseTypeDef = connection.assume_role(
```
webknjaz:
```suggestion
        response: "AssumeRoleResponseTypeDef" = connection.assume_role(
```

src/awx_plugins/credentials/aws_assumerole.py
```diff
@@ -0,0 +1,182 @@
...
+    :type secret_key: str
+    :param role_arn: The ARN received from AWS.
+    :type role_arn: str
+    :param external_id: The external ID received from AWS.
+    :type external_id: int
+    :returns: The credentials received from AWS.
+    :rtype: dict
+    :raises ValueError: If the client response is bad.
+    """
+    connection: STSClient = boto3.client(
```
webknjaz:
```suggestion
    connection: "STSClient" = boto3.client(
```

src/awx_plugins/credentials/aws_assumerole.py
```diff
@@ -0,0 +1,182 @@
...
+        AssumeRoleResponseTypeDef,
+        CredentialsTypeDef,
+    )
+
+from .plugin import CredentialPlugin
+
+
+_aws_cred_cache: dict[
+    str,
+    'CredentialsTypeDef' | dict[typing.Never, typing.Never],
```
webknjaz:
```suggestion
    'CredentialsTypeDef | dict[typing.Never, typing.Never]',
```

src/awx_plugins/credentials/aws_assumerole.py
```diff
@@ -0,0 +1,182 @@
...
+    ],
+}
+
+
+def aws_assumerole_getcreds(
+        access_key: str | None,
+        secret_key: str | None,
+        role_arn: str | None,
+        external_id: int,
+) -> 'CredentialsTypeDef' | dict[typing.Never, typing.Never]:
```
webknjaz:
```suggestion
) -> 'CredentialsTypeDef | dict[typing.Never, typing.Never]':
```

docs/conf.py
```diff
@@ -195,3 +204,43 @@
...
+    }
+    ref_target = node.get('reftarget', '')
+
+    try:
+        return reference(
+            ref_target,
+            ref_target,
+            internal=False,
+            refuri=f'https://youtype.github.io/boto3_stubs_docs/mypy_boto3_sts/{
+                boto3_type_uri_map[ref_target]}',
```
webknjaz:
```suggestion
            refuri=
            'https://youtype.github.io/boto3_stubs_docs/'
            f'mypy_boto3_sts/{boto3_type_uri_map[ref_target]}',
```

docs/conf.py
```diff
@@ -195,3 +204,43 @@
...
+            refuri='https://youtype.github.io/boto3_stubs_docs/'
+            f'mypy_boto3_sts/{boto3_type_uri_map[ref_target]}',
+        )
+    except KeyError:
+        return None
+
+
+def setup(app: Sphinx) -> dict[str, bool | str]:
+    """Register project-local Sphinx extension-API customizations."""
+
```
webknjaz:
```suggestion
    """Register project-local Sphinx extension-API customizations.
    
    :param app: Initialized Sphinx app instance.
    :type app: Sphinx

    :returns: Extension metadata.
    :rtype: dict[str, bool | str]
    """
```

src/awx_plugins/credentials/aws_assumerole.py
```diff
@@ -0,0 +1,182 @@
...
+    'required': [
+        'role_arn',
+    ],
+}
+
+
+def aws_assumerole_getcreds(
+        access_key: str | None,
+        secret_key: str | None,
+        role_arn: str | None,
```
webknjaz: This should either drop `| None` or have a check somewhere that errors out when it's `None` MyPy is unhappy about possibly passing a `None` to `assume_role()` while it clearly states that it only accepts strings.

.pre-commit-config.yaml
```diff
@@ -190,7 +190,8 @@ repos:
       @ git+https://github.com/ansible/awx_plugins.interfaces.git
     - azure-identity  # needed by credentials.azure_kv
     - azure-keyvault  # needed by credentials.azure_kv
-    - boto3-stubs  # needed by credentials.awx_secretsmanager
+    - boto3-stubs[sts]  # needed by credentials.awx_secretsmanager
```
webknjaz, line 200:
```suggestion
    - boto3-stubs[sts]  # needed by credentials.awx_secretsmanager
    - botocore-stubs  # needed by credentials.awx_secretsmanager
```

.pre-commit-config.yaml
```diff
@@ -220,7 +221,8 @@ repos:
...
```
webknjaz, line 231:
```suggestion
    - boto3-stubs[sts]  # needed by credentials.awx_secretsmanager
    - botocore-stubs  # needed by credentials.awx_secretsmanager
```

.pre-commit-config.yaml
```diff
@@ -250,7 +252,8 @@ repos:
...
```
webknjaz, line 262:
```suggestion
    - boto3-stubs[sts]  # needed by credentials.awx_secretsmanager
    - botocore-stubs  # needed by credentials.awx_secretsmanager
```

tests/credential_plugins_test.py
```diff
@@ -134,6 +135,64 @@ def test_hashivault_handle_auth_not_enough_args() -> None:
...
+
+    monkeypatch.setattr(
+        aws_assumerole,
+        'aws_assumerole_getcreds',
+        mock_getcreds,
+    )
+
+    extra_kwargs = {'identifier': identifier_key} if identifier_key else {}
+
+    token = aws_assumerole.aws_assumerole_backend(**kwargs, **extra_kwargs)
```
webknjaz:
```suggestion
    token = aws_assumerole.aws_assumerole_backend(**{**kwargs, **extra_kwargs})
```

src/awx_plugins/credentials/aws_assumerole.py
```diff
@@ -0,0 +1,182 @@
...
+
+
+if typing.TYPE_CHECKING:
+    from mypy_boto3_sts.client import STSClient
+    from mypy_boto3_sts.type_defs import (
+        AssumeRoleResponseTypeDef,
+        CredentialsTypeDef,
+    )
+
+from .plugin import CredentialPlugin
```
webknjaz, line 22: :loudspeaker: Don't worry about `pylint` not liking this line. I just disabled that check on `devel` via 0b3155efcee6ebdef456b289d57ac661e9f4a824. It is unable to handle complicated `isort` configurations.

tests/credential_plugins_test.py
```diff
@@ -134,6 +135,64 @@ def test_hashivault_handle_auth_not_enough_args() -> None:
...
+
+    monkeypatch.setattr(
+        aws_assumerole,
+        'aws_assumerole_getcreds',
+        mock_getcreds,
+    )
+
+    extra_kwargs = {'identifier': identifier_key} if identifier_key else {}
+
+    token = aws_assumerole.aws_assumerole_backend(**{**kwargs, **extra_kwargs})
```
webknjaz:
```suggestion
    token = aws_assumerole.aws_assumerole_backend(**kwargs, identifier=identifier_key or kwargs['identifier'])
```

tests/credential_plugins_test.py
```diff
@@ -134,6 +135,62 @@ def test_hashivault_handle_auth_not_enough_args() -> None:
...
+        {
+            'access_key': 'my_access_key',
+            'secret_key': 'my_secret_key',
+            'role_arn': 'the_arn',
+            'identifier': 'access_token',
+        },
+        {
+            'role_arn': 'the_arn',
+            'identifier': 'access_token',
+        },
```
webknjaz:
```suggestion
        },
        {},
```

tests/credential_plugins_test.py
```diff
@@ -134,6 +135,63 @@ def test_hashivault_handle_auth_not_enough_args() -> None:
...
+        }
+
+    monkeypatch.setattr(
+        aws_assumerole,
+        'aws_assumerole_getcreds',
+        mock_getcreds,
+    )
+
+    token = aws_assumerole.aws_assumerole_backend(
+        **kwargs, identifier=identifier_key or kwargs['identifier'])
```
webknjaz:
```suggestion
    token = aws_assumerole.aws_assumerole_backend(
        identifier=identifier_key or 'access_token',
        role_arn='the_arn',
        **kwargs,
    )
```

tests/credential_plugins_test.py
```diff
@@ -134,6 +135,61 @@ def test_hashivault_handle_auth_not_enough_args() -> None:
         hashivault.handle_auth()
 
 
+@pytest.mark.parametrize(
+    'kwargs',
```
webknjaz:
```suggestion
    'explicit_creds',
```

tests/credential_plugins_test.py
```diff
@@ -134,6 +135,61 @@ def test_hashivault_handle_auth_not_enough_args() -> None:
...
+    ),
+    (
+        (None, 'the_access_token'),
+        ('access_key', 'the_access_key'),
+        ('secret_key', 'the_secret_key'),
+    ),
+)
+def test_aws_assumerole_identifier(
+    monkeypatch: pytest.MonkeyPatch,
+    kwargs: dict[str, str], identifier_key: str | None, expected: str,
```
webknjaz:
```suggestion
    explicit_creds: dict[str, str], identifier_key: str | None, expected: str,
```

tests/credential_plugins_test.py
```diff
@@ -134,6 +135,61 @@ def test_hashivault_handle_auth_not_enough_args() -> None:
...
+    monkeypatch.setattr(
+        aws_assumerole,
+        'aws_assumerole_getcreds',
+        mock_getcreds,
+    )
+
+    token = aws_assumerole.aws_assumerole_backend(
+        identifier=identifier_key or 'access_token',
+        role_arn='the_arn',
+        **kwargs,
```
webknjaz:
```suggestion
        **explicit_creds,
```

tests/credential_plugins_test.py
```diff
@@ -134,6 +135,61 @@ def test_hashivault_handle_auth_not_enough_args() -> None:
...
 
+@pytest.mark.parametrize(
+    'explicit_creds',
+    (
+        {
+            'access_key': 'my_access_key',
+            'secret_key': 'my_secret_key',
+        },
+        {},
+    ),
```
webknjaz, line 146:
```suggestion
        ids=('with-creds-args', 'with-env-creds'),
    ),
```

tests/credential_plugins_test.py
```diff
@@ -134,6 +135,61 @@ def test_hashivault_handle_auth_not_enough_args() -> None:
...
+@pytest.mark.parametrize(
+    (
+        'identifier_key',
+        'expected',
+    ),
+    (
+        (None, 'the_access_token'),
+        ('access_key', 'the_access_key'),
+        ('secret_key', 'the_secret_key'),
+    ),
```
webknjaz, line 158:
```suggestion
    ),
    ids=(
        'access-token',
        'access-key',
        'secret-key',
    ),
```

tests/credential_plugins_test.py
```diff
@@ -134,6 +135,67 @@ def test_hashivault_handle_auth_not_enough_args() -> None:
...
+@pytest.mark.parametrize(
+    'explicit_creds',
+    (
+        {
+            'access_key': 'my_access_key',
+            'secret_key': 'my_secret_key',
+        },
+        {},
+        ids=('with-creds-args', 'with-env-creds'),
+    ),
```
webknjaz:
```suggestion
    ),
    ids=('with-creds-args', 'with-env-creds'),
```

src/awx_plugins/credentials/aws_assumerole.py
thedoubl3j: that is my thought as well, looking at the actual error message makes it seem like this should be dev focused since we are the ones going and getting this actual value with the user inputs but could be helpful for debugging if someone hits this. kinda leaves a good flag as where things could have fallen apart?

src/awx_plugins/credentials/aws_assumerole.py
thedoubl3j: I think any of these make sense and agree value isn't the correct thing in place. Only one I can see kinda not fitting is connection since errors could or could not be connection related. IMO lookup makes the most since we the actual function call is going and "lookin " something up but I am not dead set on that.

src/awx_plugins/credentials/aws_assumerole.py
thedoubl3j: done, also dropped them from the backend function as well.

src/awx_plugins/credentials/aws_assumerole.py
```diff
@@ -0,0 +1,184 @@
...
+    ],
+    'required': [
+        'role_arn',
+    ],
+}
+
+
+def aws_assumerole_getcreds(
+        access_key: str,
+        secret_key: str,
```
webknjaz: These two can still be `None` because that's what the boto3 function expects:
```suggestion
        access_key: str | None,
        secret_key: str | None,
```

src/awx_plugins/credentials/aws_assumerole.py
```diff
@@ -0,0 +1,184 @@
...
+        raise ValueError(
+            f'Got a bad client response from AWS: {client_err.message}.',
+        ) from client_err
+
+    return response.get('Credentials', {})
+
+
+def aws_assumerole_backend(
+        access_key: str,
+        secret_key: str,
```
webknjaz:
```suggestion
        access_key: str | None,
        secret_key: str | None,
```

src/awx_plugins/credentials/aws_assumerole.py
webknjaz: Technically, all the exceptions are targeting the devs, so are the tracebacks. But if awx uses this somehow, we can't change it to a `LookupError`, I think.

src/awx_plugins/credentials/aws_assumerole.py
webknjaz: Yep. That's what I often use. And the calling code would need to handle it and turn into whatever awx interface currently wants, which seems to be a `ValueError`, looking into the function below.

.pre-commit-config.yaml
```diff
@@ -190,7 +190,9 @@ repos:
...
+    - botocore-stubs  # needed by credentials.awx_secretsmanager
+    - django-stubs  # needed by credentials.injectors and inventory.plugins
```
webknjaz: bad conflict resolution
```suggestion
```

.pre-commit-config.yaml
```diff
@@ -220,7 +222,9 @@ repos:
...
```
webknjaz:
```suggestion
```

.pre-commit-config.yaml
```diff
@@ -250,7 +254,9 @@ repos:
...
```
webknjaz:
```suggestion
```

src/awx_plugins/credentials/aws_assumerole.py
```diff
@@ -0,0 +1,184 @@
...
+from datetime import datetime
+
+from awx_plugins.interfaces._temporary_private_django_api import (  # noqa: WPS436
+    gettext_noop as _,
+)
+
+import boto3
+from botocore.exceptions import ClientError
+
+from .plugin import CredentialPlugin
```
webknjaz: duplicate
```suggestion
```

tests/credential_plugins_test.py
```diff
@@ -134,6 +135,67 @@ def test_hashivault_handle_auth_not_enough_args() -> None:
...
+            'Expiration': datetime.datetime.today() + datetime.timedelta(days=1),
+        }
+
+    monkeypatch.setattr(
+        aws_assumerole,
+        'aws_assumerole_getcreds',
+        mock_getcreds,
+    )
+
+    token = aws_assumerole.aws_assumerole_backend(
```
webknjaz, line 191:
```suggestion
    token = aws_assumerole.aws_assumerole_backend(
        external_id=42,
```

tests/credential_plugins_test.py
```diff
@@ -134,6 +135,68 @@ def test_hashivault_handle_auth_not_enough_args() -> None:
...
+    monkeypatch: pytest.MonkeyPatch,
+    explicit_creds: dict[str, str], identifier_key: str | None, expected: str,
+) -> None:
+    """Test that the aws_assumerole_backend function call returns a token given
+    the access_key and secret_key."""
+
+    def mock_getcreds(
+            access_key: str | None,
+            secret_key: str | None,
+            role_arn: str | None,
```
webknjaz:
```suggestion
            role_arn: str,
```

tests/credential_plugins_test.py
```diff
@@ -134,6 +135,68 @@ def test_hashivault_handle_auth_not_enough_args() -> None:
...
+            external_id: int,
+    ) -> dict:
+        return {
+            'access_key': 'the_access_key',
+            'secret_key': 'the_secret_key',
+            'access_token': 'the_access_token',
+            'Expiration': datetime.datetime.today() + datetime.timedelta(days=1),
```
webknjaz, line 182: According to https://youtype.github.io/boto3_stubs_docs/mypy_boto3_sts/type_defs/#credentialstypedef, this should be
```suggestion
            'AccessKeyId': 'the_access_key',
            'SecretAccessKey': 'the_secret_key',
            'SessionToken': 'the_access_token',
            'Expiration': datetime.datetime.today() + datetime.timedelta(days=1),
```

@thedoubl3j do you know why the example return value diverges in tests so much?

webknjaz: (this is the reason I don't like mocking, as it just tests the stubs instead of the important bits of code)

src/awx_plugins/credentials/aws_assumerole.py
```diff
@@ -0,0 +1,182 @@
...
+    # Generate a unique SHA256 hash for combo of user access key and ARN
+    # This should allow two users requesting the same ARN role to have
+    # separate credentials, and should allow the same user to request
+    # multiple roles.
+    credential_key_hash = hashlib.sha256(
+        (str(access_key or '') + role_arn).encode('utf-8'),
+    )
+    credential_key = credential_key_hash.hexdigest()
+
+    credentials = _aws_cred_cache.get(credential_key, None)
```
webknjaz: The tests revealed another similar bug here:
```suggestion
    credentials = _aws_cred_cache.get(credential_key, {})
```
*(https://github.com/ansible/awx-plugins/actions/runs/11265198558/job/31326677813?pr=24#step:17:155)*

tests/credential_plugins_test.py
thedoubl3j: I have no idea @webknjaz all of this just pulled from the AWX PR that we had from the external contributor so reasoning etc on the decisions made I have no clue.

derekwaters: HI @webknjaz and @thedoubl3j I'm the original submitter of the PR, and you can pretty much explain any of my decisions as mistakes by a very new Python developer.

In this particular example, you're right, the actual call to boto3 should return an 'AccessKeyId', 'SecretAccessKey', 'SessionToken' 'Expiration' dict, but the mock was poorly written. Sorry about that.

I also agree that the mocking was a poor choice here, but given that the actual requirement to test that code is a valid AWS account with IAM set up, I wasn't sure of a better way (at least in the AWX project tests which didn't seem to have any of that capability) - if there is I'd love to know about it for any future work I do!

derekwaters: Also please feel free to ping me if there's anything else I can help with on this PR.

webknjaz: @derekwaters thanks! We'll keep this in mind. Having a full e2e test would be hard to achieve here and is to worth the effort. Normally, for such testing I'd go for something like a stub server to produce better coverage and test realistic behavior.
It's probably difficult to emulate in this case so the next best thing is to mock the client so that everything else runs as usual. Ideally, we should only mock things that are on the boundary between our code and third party, or something in the environment (like freezing time). These are nor particularly connected to pythonic vs. not. Unfortunately, oftentimes pythonistas don't know they can follow battle-tested architectural principles that are well-known in other ecosystems, not being exposed to their existence mostly. For example, I really like the dependency injection principle, which allows writing testable code and I believe we should design it to be testable from the very beginning, instead of testing as an afterthought and having to fight the previous decisions.

webknjaz:
@thedoubl3j this will end up in https://github.com/ansible/awx_plugins.credential.aws_assumerole eventually. That repo doesn't yet have Git history, but it will. Still, it's useful to do an interactive rebase with the commit cleanup here and fix the conflicts. This will help re-graft this branch into the new repo.

derekwaters:
Hi folks, I'm just checking in on the status of this PR? The repo reference in the last comment must be private, because it 404s for me, but I'm interested in following the progress of getting this change into AWX. Thanks!

webknjaz:
@derekwaters we didn't end up splitting the plugins repo further at the time. This is on hold for now.

pratikkumar-mohite:
Any update on this, we are still waiting this to be implemented

PR #26, opened 2024-08-30
Add type annotations to function signatures in credential_plugins_test and importable_test modules

djyasin:
This PR contains the type annotations to function signatures for credential_plugins_test.py and tests/importable_test.py

### DIFF
--- a/.mypy.ini
+++ b/.mypy.ini
@@ -34,3 +34,6 @@ strict_optional = true
 warn_no_return = true
 warn_redundant_casts = true
 warn_unused_ignores = true
+
+[mypy-awx.main.models.credential.*]  # This can be removed once AWX can be added as a dependency or eliminated for good
+ignore_missing_imports = true
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -189,6 +189,10 @@ repos:
     additional_dependencies:
     - lxml  # dep of `--txt-report`, `--cobertura-xml-report` & `--html-report`
     - pytest
+    # The following dependency is used in `credentials.aim`,
+    # `credentials.centrify_vault`, `credentials.conjur`
+    # and `credentials.hashivault`:
+    - types-requests
     args:
     - --python-version=3.13
     - --txt-report=.tox/.tmp/.test-results/mypy--py-3.13
@@ -201,6 +205,10 @@ repos:
     additional_dependencies:
     - lxml  # dep of `--txt-report`, `--cobertura-xml-report` & `--html-report`
     - pytest
+    # The following dependency is used in `credentials.aim`,
+    # `credentials.centrify_vault`, `credentials.conjur`
+    # and `credentials.hashivault`:
+    - types-requests
     args:
     - --python-version=3.12
     - --txt-report=.tox/.tmp/.test-results/mypy--py-3.12
@@ -213,6 +221,10 @@ repos:
     additional_dependencies:
     - lxml  # dep of `--txt-report`, `--cobertura-xml-report` & `--html-report`
     - pytest
+    # The following dependency is used in `credentials.aim`,
+    # `credentials.centrify_vault`, `credentials.conjur`
+    # and `credentials.hashivault`:
+    - types-requests
     args:
     - --python-version=3.11
     - --txt-report=.tox/.tmp/.test-results/mypy--py-3.11
--- a/tests/credential_plugins_test.py
+++ b/tests/credential_plugins_test.py
@@ -5,7 +5,7 @@ import pytest
 from awx_plugins.credentials import hashivault
 
 
-def test_imported_azure_cloud_sdk_vars():
+def test_imported_azure_cloud_sdk_vars() -> None:
     from awx_plugins.credentials import azure_kv
 
     assert len(azure_kv.clouds) > 0
@@ -14,7 +14,7 @@ def test_imported_azure_cloud_sdk_vars():
     assert all([hasattr(c.suffixes, 'keyvault_dns') for c in azure_kv.clouds])
 
 
-def test_hashivault_approle_auth():
+def test_hashivault_approle_auth() -> None:
     kwargs = {
         'role_id': 'the_role_id',
         'secret_id': 'the_secret_id',
@@ -27,7 +27,7 @@ def test_hashivault_approle_auth():
     assert res == expected_res
 
 
-def test_hashivault_kubernetes_auth():
+def test_hashivault_kubernetes_auth() -> None:
     kwargs = {
         'kubernetes_role': 'the_kubernetes_role',
     }
@@ -44,7 +44,7 @@ def test_hashivault_kubernetes_auth():
         assert res == expected_res
 
 
-def test_hashivault_client_cert_auth_explicit_role():
+def test_hashivault_client_cert_auth_explicit_role() -> None:
     kwargs = {
         'client_cert_role': 'test-cert-1',
     }
@@ -55,8 +55,8 @@ def test_hashivault_client_cert_auth_explicit_role():
     assert res == expected_res
 
 
-def test_hashivault_client_cert_auth_no_role():
-    kwargs = {}
+def test_hashivault_client_cert_auth_no_role() -> None:
+    kwargs: dict[str, str] = {}
     expected_res = {
         'name': None,
     }
@@ -64,14 +64,14 @@ def test_hashivault_client_cert_auth_no_role():
     assert res == expected_res
 
 
-def test_hashivault_userpass_auth():
+def test_hashivault_userpass_auth() -> None:
     kwargs = {'username': 'the_username', 'password': 'the_password'}
     expected_res = {'username': 'the_username', 'password': 'the_password'}
     res = hashivault.userpass_auth(**kwargs)
     assert res == expected_res
 
 
-def test_hashivault_handle_auth_token():
+def test_hashivault_handle_auth_token() -> None:
     kwargs = {
         'token': 'the_token',
     }
@@ -79,7 +79,7 @@ def test_hashivault_handle_auth_token():
     assert token == kwargs['token']
 
 
-def test_hashivault_handle_auth_approle():
+def test_hashivault_handle_auth_approle() -> None:
     kwargs = {
         'role_id': 'the_role_id',
         'secret_id': 'the_secret_id',
@@ -91,7 +91,7 @@ def test_hashivault_handle_auth_approle():
         assert token == 'the_token'
 
 
-def test_hashivault_handle_auth_kubernetes():
+def test_hashivault_handle_auth_kubernetes() -> None:
     kwargs = {
         'kubernetes_role': 'the_kubernetes_role',
     }
@@ -110,7 +110,7 @@ def test_hashivault_handle_auth_kubernetes():
             assert token == 'the_token'
 
 
-def test_hashivault_handle_auth_client_cert():
+def test_hashivault_handle_auth_client_cert() -> None:
     kwargs = {
         'client_cert_public': 'foo',
         'client_cert_private': 'bar',
@@ -126,7 +126,7 @@ def test_hashivault_handle_auth_client_cert():
         assert token == 'the_token'
 
 
-def test_hashivault_handle_auth_not_enough_args():
+def test_hashivault_handle_auth_not_enough_args() -> None:
     with pytest.raises(Exception):
         hashivault.handle_auth()
 
@@ -137,12 +137,12 @@ class TestDelineaImports:
     library, so these tests are designed to fail if these wind up using the
     fallback import."""
 
-    def test_dsv_import(self):
+    def test_dsv_import(self) -> None:
         from awx_plugins.credentials.dsv import SecretsVault  # noqa
         # assert this module as opposed to older thycotic.secrets.vault
         assert SecretsVault.__module__ == 'delinea.secrets.vault'
 
-    def test_tss_import(self):
+    def test_tss_import(self) -> None:
         from awx_plugins.credentials.tss import DomainPasswordGrantAuthorizer, PasswordGrantAuthorizer, SecretServer, ServerSecret  # noqa
 
         for cls in (
--- a/tests/importable_test.py
+++ b/tests/importable_test.py
@@ -16,7 +16,7 @@ class EntryPointParam:
     name: str
     spec: str
 
-    def __str__(self):
+    def __str__(self) -> str:
         """Render an entry-point parameter as a string.
 
         To be used as a part of parametrized test ID.
@@ -159,7 +159,7 @@ with_all_plugins = pytest.mark.parametrize(
 
 
 @with_all_plugins
-def test_entry_points_exposed(entry_point: str) -> None:
+def test_entry_points_exposed(entry_point: EntryPointParam) -> None:
     """Verify the plugin entry points are discoverable.
 
     This check relies on the plugin-declaring distribution package to be
@@ -172,7 +172,9 @@ def test_entry_points_exposed(entry_point: str) -> None:
 
 
 @with_credential_plugins
-def test_entry_points_are_credential_plugin(entry_point: str) -> None:
+def test_entry_points_are_credential_plugin(
+        entry_point: EntryPointParam,
+) -> None:
     """Ensure all exposed credential plugins are of the same class."""
     entry_points = _discover_entry_points(group=entry_point.group)
     loaded_plugin_class = entry_points[entry_point.name].load()
@@ -182,7 +184,9 @@ def test_entry_points_are_credential_plugin(entry_point: str) -> None:
 
 
 @with_inventory_plugins
-def test_entry_points_are_inventory_plugin(entry_point: str) -> None:
+def test_entry_points_are_inventory_plugin(
+        entry_point: EntryPointParam,
+) -> None:
     """Ensure all exposed inventory plugins are of the same class."""
     entry_points = _discover_entry_points(group=entry_point.group)
     loaded_plugin_class = entry_points[entry_point.name].load()


webknjaz:
Judging by the RTD failures, the branch needs to be rebased (and some CI jobs will become green as a result).

webknjaz:
@djyasin here's the current state of MyPy coverage: https://app.codecov.io/gh/ansible/awx-plugins?flags%5B0%5D=MyPy. You can look into that or check what's in `python3 -Im webbrowser $(ls -1 .tox/.tmp/.test-results/mypy--py-*/index.html | head -n 1)` after each time you re-run `tox r -e pre-commit -q -- mypy --all-files` locally.

src/awx_plugins/credentials/aim.py
```diff
@@ -1,6 +1,6 @@
 from urllib.parse import quote, urlencode, urljoin
 
-import requests
+import requests  # type: ignore[import-not-found]
```
webknjaz: This should be added to https://github.com/ansible/awx-plugins/blob/472fb6f/.pre-commit-config.yaml#L189-L191 instead of ignoring.

tests/credential_plugins_test.py
```diff
@@ -137,13 +139,13 @@ class TestDelineaImports:
     library, so these tests are designed to fail if these wind up using the
     fallback import."""
 
-    def test_dsv_import(self):
-        from awx_plugins.credentials.dsv import SecretsVault  # noqa
+    def test_dsv_import(self) -> None:
+        from awx_plugins.credentials.dsv import SecretsVault  # type: ignore[import-not-found] # noqa
```
webknjaz: these ignores don't seem justified

.pre-commit-config.yaml
```diff
@@ -189,6 +189,7 @@ repos:
     additional_dependencies:
     - lxml  # dep of `--txt-report`, `--cobertura-xml-report` & `--html-report`
     - pytest
+    - requests
```
webknjaz: Please add comments next to each line, explaining what uses the deps, just like in pylint down below.

*(https://github.com/ansible/awx-plugins/blob/82c8c95/.pre-commit-config.yaml#L227-L238)*

tests/importable_test.py
```diff
@@ -185,6 +185,6 @@ def test_entry_points_are_credential_plugin(entry_point: str) -> None:
 def test_entry_points_are_inventory_plugin(entry_point: str) -> None:
```
webknjaz: I think I missed this in between rounds of refactoring the PR that introduced the tests. The actual argument type in all the tests is `EntryPointParam`. “Parametrize” used to be structured differently and inject strings into tests, before I changed the data struct but forgot to replace `str` here...

.pre-commit-config.yaml
djyasin: Thank you so much for the heads up about this one! Please let me know if what I've done looks correct. Pytest is a bit different than the others, so just let me know if anything needs to be revised.

djyasin:
This PR contains all of the fixes for `credential_plugins_test.py` and `tests/importable_test.py`

.pre-commit-config.yaml
```diff
@@ -188,7 +188,8 @@ repos:
     name: MyPy, for Python 3.13
     additional_dependencies:
     - lxml  # dep of `--txt-report`, `--cobertura-xml-report` & `--html-report`
-    - pytest
+    - pytest # dep of `credential_plugins_test.py`& `importable_test.py`
+    - requests # dep of `credentials.aim`, `credentials.centrify_vault`, `credentials.conjur` & `credentials.hashivault`
```
webknjaz: https://github.com/ansible/awx-plugins/actions/runs/10740379607/job/29788420674?pr=26#step:18:790 revealed that `requests` doesn't have typing information bundled in its dists, apparently. Sometimes, projects have separate type stubs (`*.pyi` files) published as their own PyPI projects. In this case, install them instead, for type-checking purposes:
```suggestion
    - types-requests  # dep of `credentials.aim`, `credentials.centrify_vault`, `credentials.conjur` & `credentials.hashivault`
```

webknjaz: Note the annotations that got posted here from GHA — they are coming from the linters and I see they mention needing two spaces before hashes, when adding code comments.

.pre-commit-config.yaml
```diff
@@ -188,7 +188,8 @@ repos:
     name: MyPy, for Python 3.13
     additional_dependencies:
     - lxml  # dep of `--txt-report`, `--cobertura-xml-report` & `--html-report`
-    - pytest
+    - pytest # dep of `credential_plugins_test.py`& `importable_test.py`
```
webknjaz: We run MyPy 3 times, against different Python versions — each of the invocations must have the same list of dependencies copied over, not just the Python 3.13 run.

webknjaz:
This problem https://github.com/ansible/awx-plugins/actions/runs/10740379607/job/29788419764?pr=26#step:18:29 shows that there is a dependency on `awx`:
```console
src/awx_plugins/credentials/plugins.py:7:1: error: Cannot find implementation
or library stub for module named "awx.main.models.credential" 
[import-not-found]
```

Though, we cannot physically declare it due to it not being on PyPI (well, there's a wrong `awx` there but not ours). So we have no choice but to ignore it since we can't check against it until we decouple the two projects, exterminating the circular dependencies between them.

The way to do it is by reproducing a section like https://github.com/sphinx-contrib/sphinxcontrib-towncrier/blob/b01f3b0/.mypy.ini#L36-L40 and configuring just that section to pretend there are no such imports...

tests/credential_plugins_test.pyi
webknjaz, line 1: It is not necessary to add type stub files. They are used instead of normal `.py` modules. But if those already have typing, you're overshadowing the respective information for no good reason. Type stubs exist to have fully disconnected typing, which used to be useful for projects supporting both Python 2 and 3 at the same time, for example. This case is different since we're not bound by said limitation and can just put types into regular source code.

tests/importable_test.pyi
```diff
@@ -0,0 +1,21 @@
...
+
+
+@dataclass(frozen=True)
+class EntryPointParam:
+    def __init__(self, group: str, name: str, spec: str) -> None:
+        self.group = group
+        self.name = name
+        self.spec = spec
+
+credential_plugins: Incomplete
```
webknjaz: Why do you think it's incomplete? MyPy is perfectly capable of inferring the types the way they are set. Anyway, I don't think `*.pyi` are needed, as I mentioned in the other comment.

.mypy.ini
```diff
@@ -34,3 +34,9 @@ strict_optional = true
 warn_no_return = true
 warn_redundant_casts = true
 warn_unused_ignores = true
+
+[mypy-towncrier.*]
+ignore_missing_imports = true
+
+[mypy-sphinxcontrib.towncrier.*]
+ignore_missing_imports = true
```
webknjaz: These were just examples from another project. We don't have these deps currently. So we don't need to ignore them as they're never encountered since the imports don't exist.

.pre-commit-config.yaml
```diff
@@ -188,7 +188,8 @@ repos:
     name: MyPy, for Python 3.13
     additional_dependencies:
     - lxml  # dep of `--txt-report`, `--cobertura-xml-report` & `--html-report`
-    - pytest
+    - pytest  # dep of `credential_plugins_test.py`& `importable_test.py`
```
webknjaz: I think that `pytest` does not need explanation since it's a commonly used framework that will be in each test file. It'd be silly to enumerate like 50 test modules when the tests grow. Either don't explain `pytest` at all or just summarize that it's used in tests and MyPy lints those files too.

webknjaz: I initially didn't add that comment because I don't think it needs one, reserving the comments for less obvious deps. For example, just by seeing `lxml` for the first time, it would never occur to me that it makes certain specific MyPy CLI flags work.

webknjaz:
FYI, this failure https://github.com/ansible/awx-plugins/actions/runs/10818045643/job/30012894850?pr=26#step:18:49 is flaky. I reported it upstream via https://github.com/sphinx-contrib/spelling/issues/227. For now, just restart the job if you see this error.

.mypy.ini
```diff
@@ -34,3 +34,6 @@ strict_optional = true
 warn_no_return = true
 warn_redundant_casts = true
 warn_unused_ignores = true
+
+[mypy-awx.main.models.credential.*] # This can be removed once AWX has been imported
```
webknjaz: It's customary to have double whitespaces before comments in almost any language, so it's best to keep that habit even where it's not linted, for consistency.
```suggestion
[mypy-awx.main.models.credential.*]  # This can be removed once AWX has been imported
```

.pre-commit-config.yaml
```diff
@@ -189,6 +189,8 @@ repos:
     additional_dependencies:
     - lxml  # dep of `--txt-report`, `--cobertura-xml-report` & `--html-report`
     - pytest
+    - requests  # dep of `credentials.aim`, `credentials.centrify_vault`,
+    # `credentials.conjur` & `credentials.hashivault`
```
webknjaz: I've never seen comments added below lines they refer to. It's always either above or on the same line. The same line should be preferred, since it guards against new things appearing between the comment and the code, making them drift away from each other over time. The preceding line is the last resort in cases where we can't satisfy the line length expectations.
```suggestion
    # The following dependency is used in `credentials.aim`,
    # `credentials.centrify_vault`, `credentials.conjur`
    # and `credentials.hashivault`:
    - requests
```

.pre-commit-config.yaml
```diff
@@ -201,6 +203,8 @@ repos:
...
```
webknjaz:
```suggestion
    # The following dependency is used in `credentials.aim`,
    # `credentials.centrify_vault`, `credentials.conjur`
    # and `credentials.hashivault`:
    - requests
```

.pre-commit-config.yaml
```diff
@@ -213,6 +217,8 @@ repos:
...
```
webknjaz:
```suggestion
    # The following dependency is used in `credentials.aim`,
    # `credentials.centrify_vault`, `credentials.conjur`
    # and `credentials.hashivault`:
    - requests
```

tests/credential_plugins_test.py
```diff
@@ -137,13 +137,19 @@ class TestDelineaImports:
     library, so these tests are designed to fail if these wind up using the
     fallback import."""
 
-    def test_dsv_import(self):
-        from awx_plugins.credentials.dsv import SecretsVault  # noqa
+    def test_dsv_import(self) -> None:
+        from awx_plugins.credentials.dsv import SecretsVault  # noqa: F401
```
webknjaz: This is not related to MyPy/typing. Could you move the change out into a separate PR? That PR will have to also make the CI changes since this update makes https://github.com/ansible/awx-plugins/actions/runs/10818045643/job/30020907062?pr=26 pass, and it needs to move back into the regular pre-commit job.

tests/credential_plugins_test.py
```diff
@@ -137,13 +137,19 @@ class TestDelineaImports:
...
 
-    def test_tss_import(self):
-        from awx_plugins.credentials.tss import DomainPasswordGrantAuthorizer, PasswordGrantAuthorizer, SecretServer, ServerSecret  # noqa
+    def test_tss_import(self) -> None:
+        from awx_plugins.credentials.tss import (  # noqa: F401
+            DomainPasswordGrantAuthorizer,
+            PasswordGrantAuthorizer,
+            SecretServer,
+            ServerSecret,
+        )
```
webknjaz: Same as above, let's move this out to keep the PR atomic.

webknjaz (changes requested):
Let's move the unrelated fixes elsewhere so that this PR won't have to accumulate more changes required for those unrelated fixes to be complete.

webknjaz:
@djyasin I just realized that the PR description and title need to be updated to reflect what's fixed. And also to be accurate...

.mypy.ini
```diff
@@ -34,3 +34,6 @@ strict_optional = true
 warn_no_return = true
 warn_redundant_casts = true
 warn_unused_ignores = true
+
+[mypy-awx.main.models.credential.*]  # This can be removed once AWX has been imported
```
webknjaz:
```suggestion
[mypy-awx.main.models.credential.*]  # This can be removed once AWX can be added as a dependency or eliminated for good
```

.pre-commit-config.yaml
djyasin: This has been resolved. I have removed the comments around Pytests in all Python versions:
.pre-commit-config.yaml

tests/credential_plugins_test.py
djyasin: I have removed those codes and have reverted both of these comments back to `# noqa`.

tests/credential_plugins_test.py
djyasin: I have removed those codes and have reverted both of these comments back to # noqa.

.pre-commit-config.yaml
djyasin: Please let me know if I didn't resolve this correctly!

.pre-commit-config.yaml
```diff
@@ -189,6 +189,11 @@ repos:
     additional_dependencies:
     - lxml  # dep of `--txt-report`, `--cobertura-xml-report` & `--html-report`
     - pytest
+    # The following dependency is used in `credentials.aim`,
+    # `credentials.centrify_vault`, `credentials.conjur`
+    # and `credentials.hashivault`:
+    - types-requests  # dep of `credentials.aim`, `credentials.centrify_vault`,
+    # `credentials.conjur` & `credentials.hashivault`
```
webknjaz: No need to repeat the above comment once more.
```suggestion
    - types-requests
```

.pre-commit-config.yaml
```diff
@@ -201,6 +206,11 @@ repos:
...
```
webknjaz:
```suggestion
    - types-requests
```

.pre-commit-config.yaml
```diff
@@ -213,6 +223,11 @@ repos:
...
```
webknjaz:
```suggestion
    - types-requests
```

tests/credential_plugins_test.py
```diff
@@ -137,13 +137,18 @@ class TestDelineaImports:
...
 
-    def test_tss_import(self):
-        from awx_plugins.credentials.tss import DomainPasswordGrantAuthorizer, PasswordGrantAuthorizer, SecretServer, ServerSecret  # noqa
+    def test_tss_import(self) -> None:
+        from awx_plugins.credentials.tss import (  # noqa
+            DomainPasswordGrantAuthorizer,
+            PasswordGrantAuthorizer,
+            SecretServer,
+            ServerSecret,
+        )
```
webknjaz: Plz keep the original formatting. Formatting changes must never be mixed with the functional ones, due to the confusing perception this causes by making it look like these lines have *something* related to the goal advertised in the PR title, when there's none.

webknjaz (changes requested):
There's a few places that need to be addressed still.

tests/credential_plugins_test.py
webknjaz: Looks like the revert was partial and needs to be iterated on: https://github.com/ansible/awx-plugins/pull/26#discussion_r1759175316.

Also, it's a good idea to put this change into a new PR and paste a link to that PR here.

tests/credential_plugins_test.py
djyasin: Ah, okay! Thank you!

tests/credential_plugins_test.py
djyasin: I'm so sorry, I am a little confused about what you mean exactly. Could you be more explicit?

tests/credential_plugins_test.py
```diff
@@ -137,14 +137,13 @@ class TestDelineaImports:
...
 
-    def test_dsv_import(self):
+    def test_dsv_import(self) -> None:
         from awx_plugins.credentials.dsv import SecretsVault  # noqa
         # assert this module as opposed to older thycotic.secrets.vault
         assert SecretsVault.__module__ == 'delinea.secrets.vault'
 
-    def test_tss_import(self):
+    def test_tss_import(self) -> None:
         from awx_plugins.credentials.tss import DomainPasswordGrantAuthorizer, PasswordGrantAuthorizer, SecretServer, ServerSecret  # noqa
```
webknjaz, line 146: Keep that original separator line
```suggestion
        from awx_plugins.credentials.tss import DomainPasswordGrantAuthorizer, PasswordGrantAuthorizer, SecretServer, ServerSecret  # noqa

```

tests/credential_plugins_test.py
webknjaz: @djyasin those noqa changes were good but misplaced. In order for us not to lose a necessary and good patch, you should move that patch into a separate branch and open a PR with it. That work still needs to be done. So instead of disregarding it from Git entirely, copy it into a separate PR and iterate on fixing `python-check-blanket-noqa` there.

.pre-commit-config.yaml
webknjaz: Marking this thread as resolved.

.pre-commit-config.yaml
webknjaz: Marking this thread as resolved.

tests/credential_plugins_test.py
djyasin: Thank you so much, that makes sense! I've done that over here: https://github.com/ansible/awx-plugins/pull/36

webknjaz (approved):
@djyasin feel free to squash the commits and put this into the merge queue.

tests/credential_plugins_test.py
webknjaz: Marking this as resolved.

tests/credential_plugins_test.py
webknjaz: Marking this as resolved.

webknjaz:
@djyasin I think the PR title needs to be clarified still. It's pretty generic in describing what is done. It would be nice to see what exactly is being changed.

Also, some MyPy violations going away is one of the effects of the patch. The title should be talking about the change that leads to those effects. Like adding typing. Or rather adding type annotations to function signatures in these modules. Not sure if this covers all violations in these modules (probably not?)

tests/importable_test.py
webknjaz, line 1: https://github.com/ansible/awx-plugins/actions/runs/10854240324/job/30124294717?pr=26#step:18:1029 revealed there's one missing type annotation left in this module:
```console
tests/importable_test.py:19:5: error: Function is missing a type annotation 
[no-untyped-def]
        def __str__(self):
        ^
```

djyasin: a30e167bd2762fae11b693a15fead4554f1f8825 this commit contains the resolution for this one

webknjaz (approved):

PR #27, opened 2024-09-04
Move inventory and credential plugins docs to `awx-plugins`

tvo318:
Moved credential plugins and inventory plugins (inventory source) chapters from the awx user guide.

Previews:
* https://awx-plugins-core--27.org.readthedocs.build/en/27/credential-plugins/
* https://awx-plugins-core--27.org.readthedocs.build/en/27/inventory-plugins/
* https://awx-plugins-core--27.org.readthedocs.build/en/27/inventory-plugins-templates/


.obsidian/app.json
webknjaz, line 1: Could you avoid including your editor configuration in this PR?

Here are some pointers on how to configure that:
* https://docs.github.com/en/get-started/getting-started-with-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer
* https://stackoverflow.com/questions/7335420/global-git-ignore
* https://git-scm.com/docs/gitignore

docs/index.md
```diff
@@ -21,15 +21,18 @@
...
 
+credential-plugins
+inventory-plugins
+inventory-plugins-templates
 changelog
 ```
 
 ```{toctree}
 :caption: Contributing
 :hidden: true
```
webknjaz: let's keep this visual separator
```suggestion
:hidden: true

```

docs/contributing/communication.md
webknjaz, line 1: Could you submit this document separately, for now?

docs/inventory-plugins-templates.rst
```diff
@@ -0,0 +1,413 @@
...
+
+Amazon Web Services EC2
+------------------------
+
+.. index::
+   pair: inventories; Amazon Web Services
+   pair: inventories; aws
+   pair: inventory plugins; aws
+
+::
```
webknjaz, line 30: Since this PR is about moving code from one place to another, let's keep this. But a nice follow-up could be adding code highlighting, maybe.

docs/contributing/communication.md
webknjaz: Did you mean to keep this converted to Markdown, by the way?

webknjaz:
It occurred to me that it might be a good idea to preserve some of that Git history related to the moved files. I'll see if that's possible.

webknjaz:
@tvo318 please use https://github.com/ansible/awx-plugins/actions/runs/10694993639?pr=27 to see if the patch is good enough. The goal is to have all the jobs that have `-docs` in their names green. Currently, I see that the linkcheck one caught a dead link. And the spellcheck one needs to be handled by either correcting the words or adding them to https://github.com/ansible/awx-plugins/blob/472fb6f/docs/spelling_wordlist.txt.

docs/contributing/communication.md
tvo318: Yes this was in markdown. Is that ok?

.obsidian/app.json
tvo318: I didn't use Obsidian, I used Sublime. Why did those configurations show up?

docs/contributing/communication.md
webknjaz: Yeah, it's fine. It's just that I saw https://github.com/ansible/awx/blob/devel/docs/docsite/rst/contributor/communication.rst, and you mentioned wanting to go back to RST for files that were pre-existing in that format and thought that you maybe forgot to do it with this one.

.obsidian/app.json
webknjaz: @tvo318 something on your computer put them there, and then you did a `git add` and `git commit`, including them here. I don't even know what that tool is :)

webknjaz:
@tvo318 I also noticed that some pre-commit checks also fail on spell checking: https://results.pre-commit.ci/run/github/836873755/1725423346.FiRHwKOaQIW764amIn0hTQ. So that's something else to look into. Many of those checks are failing for unrelated reasons but the spell check is related to your PR.

.obsidian/app.json
tvo318: OK, I'll add them to my local .gitignore file and make sure they don't get checked in on the next commit. Thanks!

docs/contributing/communication.md
tvo318: It was easy enough to keep it converted, so I left it alone.

.obsidian/app.json
webknjaz: You mean the global one at `~/.config/git/ignore`, right? Not the local in-repo config.

docs/contributing/communication.md
webknjaz: Okay, it's fine with me. Just wanted to double-check.

docs/index.md
```diff
@@ -41,6 +45,7 @@ Private unsupported (dev) API autodoc <pkg/awx_plugins.credentials>
 contributing/release_guide
 ```
 
+
```
webknjaz: Let's not add extra new lines in places that aren't related to this effort.
```suggestion
```

.obsidian/app.json
tvo318: No one else should be generating these files, so I didn't add them to the global one.

webknjaz: @tvo318 `~/.config/git/ignore` is your computer-global ignore config, not the in-repo one. You should put these exclusions there.

webknjaz:
This will need to be rebased to exclude the commits that add and remove unnecessary files. I'll probably do that myself once the content is ready.

docs/spelling_wordlist.txt
```diff
@@ -4,3 +4,4 @@ Ovirt
 Pre
 Submodules
 Subpackages
+Hashi
```
webknjaz: To keep the words sorted in this file, let's move it before `hasn`.

.gitignore
```diff
@@ -234,6 +234,9 @@ cython_debug/
 
 # Sphinx documentation
 
+# Editors
```
webknjaz: Don't add these entries to the repo because they are not related to the project.

https://seankegel.com/git-quick-tip-using-a-global-gitignore

Instead, put these lines into a file located at `~/.config/git/ignore` on your computer (this is a path that exists outside the repository checkout; if it doesn't — run `mkdir -pv ~/.config/git/ && code ~/.config/git/ignore`).

Here's an example of such a file: https://github.com/mikeslattery/dotfiles/blob/master/.config/git/ignore.

samccann:
Some files are .md and some are rst - What is the ultimate plan for docs in this repo?

webknjaz:
> Some files are .md and some are rst - What is the ultimate plan for docs in this repo?

@samccann I integrated `MyST-parser` from the very beginning. So the infrastructure supports both MyST markdown and RST. This means that we don't have to choose one or the other and both can co-exist very well. If a contributor decides to submit a Markdown document, there is no compromise on the functionality. Yet, we can keep using RST where it feels right or if we already have pre-existing documents.

chrismeyersfsu:
Content looks good to me.

chrismeyersfsu (approved):

samccann:
my recommendation is to remove these files from the docs build. We don't yet know how it all applies, so I would err on the side of keeping them in the repo only and not part of the awx-plugin docsite for now.

webknjaz:
@samccann I don't think it's a good idea to keep files that do nothing and aren't linted lying around in the repo. We can remove them later if that's what's needed.

webknjaz:
FTR here's like this bit of history got grafted into the existing Git tree: https://asciinema.org/a/675613.

samccann:
@webknjaz - I disagree. Your repo, your rules, but surfacing user documentation when we haven't decided whether there will be any gui-based user docs in this repo seems misleading imo.

chrismeyersfsu (approved):

tvo318:
@samccann - we aren't announcing this repo yet, as far as I know. We are still telling the community this is in transition and is subject to change. Right now, the focus is to move content from the awx repo. This is accomplished by this PR. We will open a new PR to clean up this content once we know more about how these plugins are supported (UI, API, etc). In the meantime, we will have it in place so making updates to the docs will be a lot easier.

webknjaz:
@samccann I hear you, I was just pointing out that the PR wouldn't even be mergeable without integrations. Also, if the content shouldn't be here, it should be deleted rather than remain in the repo. That's someone else's call, though.

PR #35, opened 2024-09-12
🎨 Add Ansible brand colors to the docs theme

samccann:
This uses the Ansible pool logo for a splash of color, and a different color for the left-hand navigation to make it stand out a bit more.

Preview: https://awx-plugins-core--35.org.readthedocs.build.


webknjaz:
@samccann do you think these are all the changes that would be necessary in a new theme?

I checked the color contrast with DevTools (against the background) and it's good for the dark theme variant but not ideal for the light one. Do we have a darker logo variant that would satisfy the accessibility recommendations?

docs/conf.py
```diff
@@ -92,9 +92,14 @@
 
 html_static_path = ['_static']
 html_theme = 'furo'
+html_logo = '_static/images/Ansible-Mark-RGB_Pool.svg'
 html_theme_options = {
-    'dark_logo': 'images/Ansible-Mark-RGB_White.svg',
-    'light_logo': 'images/Ansible-Mark-RGB_Black.svg',
+    'light_css_variables': {
+        'color-background-secondary': '#dee3ed',  # for navigation + ToC
```
webknjaz, line 98: So the only contrasting color to this I've found was `#414141`. But it's dark, so it's not a good fit for the light theme. This means that the logo color needs to be darker to have a good enough contrast.

samccann: afaik there is no other logo variant other than the black/white ones you had here already.

samccann:
So the goal here was to make as few changes as possible for now. There could be more work if we really want to customize this as he theme has a lot of flexibility on playing with colors.  I thought the Ansible pool logo had the splash of Ansible color that people could recognize.

As for the nav/toc change - I found the allwhite navigation harder to notice so tried to make that a different color that still worked with existing font colors etc.

docs/conf.py
webknjaz: Looks like https://github.com/ansible/logos/ has a few. I think I saw gray, mango, pride. Do you think any of those would work with the light theme?

webknjaz:
@samccann I also noticed that the logo size got smaller. Is that on purpose?

webknjaz:
So Firefox's accessibility tool reports contrast problems with the ToC titles in the light theme variant. It needs to be fixed before the change is acceptable UX-wise. I'd like to also compare with what Lighthouse says.

samccann:
Logo size - yes that was deliberate. The existing logo imo is too large. That said, since this is awx-related, do we want to use the [awx logo](https://github.com/ansible/logos/blob/main/awx/awx-logo.png)?

webknjaz:
That's a good question. Personally I liked the original black-and-white variants. Anyway, I don't think I should be deciding which logo to use beyond checking the accessibility implications that I'm doing.

cc @chrismeyersfsu @TheRealHaoLiu @thedoubl3j @djyasin plz share your thoughts on this

samccann:
Going to close this PR out for now. I won't get back to it for some time and think I was just bringing my own bias (aka what I was used to looking at) into this theme. My nickel would be the team should decide on whether the logo is AWX or Ansible as the only change to make.

PR #36, opened 2024-09-13
🧪💅 Eliminate blanket `# noqa` comments

djyasin:
This PR is to address the changes needed for `python-check-blanket-noqa`.

### DIFF
--- a/tests/credential_plugins_test.py
+++ b/tests/credential_plugins_test.py
@@ -138,12 +138,18 @@ class TestDelineaImports:
     fallback import."""
 
     def test_dsv_import(self) -> None:
-        from awx_plugins.credentials.dsv import SecretsVault  # noqa
+        from awx_plugins.credentials.dsv import SecretsVault  # noqa: F401
+
         # assert this module as opposed to older thycotic.secrets.vault
         assert SecretsVault.__module__ == 'delinea.secrets.vault'
 
     def test_tss_import(self) -> None:
-        from awx_plugins.credentials.tss import DomainPasswordGrantAuthorizer, PasswordGrantAuthorizer, SecretServer, ServerSecret  # noqa
+        from awx_plugins.credentials.tss import (  # noqa: F401
+            DomainPasswordGrantAuthorizer,
+            PasswordGrantAuthorizer,
+            SecretServer,
+            ServerSecret,
+        )
 
         for cls in (
                 DomainPasswordGrantAuthorizer,
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -605,7 +605,7 @@ jobs:
         - spellcheck-docs
         environment-variables:
         - >-  # only affects pre-commit, set for all for simplicity:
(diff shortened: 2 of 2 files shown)


tests/credential_plugins_test.py
```diff
@@ -137,13 +137,19 @@ class TestDelineaImports:
...
         assert SecretsVault.__module__ == 'delinea.secrets.vault'
 
     def test_tss_import(self):
-        from awx_plugins.credentials.tss import DomainPasswordGrantAuthorizer, PasswordGrantAuthorizer, SecretServer, ServerSecret  # noqa
+        from awx_plugins.credentials.tss import (  # noqa: F401
+            DomainPasswordGrantAuthorizer,
+            PasswordGrantAuthorizer,
+            SecretServer,
+            ServerSecret,
+        )
```
webknjaz: @chrismeyersfsu do these tests actually make sense in this form with in-function imports? Other smoke tests already import them, so that would influence the in-process module cache anyway.

tests/credential_plugins_test.py
```diff
@@ -137,13 +137,19 @@ class TestDelineaImports:
     library, so these tests are designed to fail if these wind up using the
     fallback import."""
 
-    def test_dsv_import(self):
-        from awx_plugins.credentials.dsv import SecretsVault  # noqa
+    def test_dsv_import(self) -> None:
+        from awx_plugins.credentials.dsv import SecretsVault  # noqa: F401
```
webknjaz, line 141: @djyasin so the best action here is to try to actually fix the underlying problem. I wonder if it's okay for us to do a `pytest.importorskip()` (if this test is even needed / cc @chrismeyersfsu).

If that's not acceptable for some reason, let's keep it as it is in your current patch.

tests/credential_plugins_test.py
```diff
@@ -137,13 +137,19 @@ class TestDelineaImports:
     library, so these tests are designed to fail if these wind up using the
     fallback import."""
 
-    def test_dsv_import(self):
-        from awx_plugins.credentials.dsv import SecretsVault  # noqa
+    def test_dsv_import(self) -> None:
```
webknjaz, line 140: This bit belongs in #26. But that's okay: it'll disappear when you rebase this PR.

webknjaz (commented):
@djyasin so since https://github.com/ansible/awx-plugins/actions/runs/10854294898/job/30125329367?pr=36 is green, this job needs to be removed and the check re-integrated into the unified pre-commit one.

More specifically, this means that the following bits should be removed: https://github.com/ansible/awx-plugins/blob/96fdebc/.github/workflows/ci-cd.yml#L616-L622 / https://github.com/ansible/awx-plugins/blob/96fdebc/.github/workflows/ci-cd.yml#L608C34-L608C60.

tests/credential_plugins_test.py
webknjaz: So now you need to resolve a conflict when rebasing.

webknjaz: I did that. Marking as resolved.

webknjaz (approved):

PR #47, opened 2024-10-16
Hide GET params that contain senstive data from error output

thedoubl3j:
Currently, the AIM plugin is showing AppID, Object_Query and reason in the constructed URL when an error is throw and streamed back to stdout. This PR will hide those GET params from being displayed.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,7 +18,7 @@ dependencies = [  # runtime deps  # https://packaging.python.org/en/latest/guide
   # GUIDANCE: only add things that this project imports directly
   # GUIDANCE: only set lower version bounds
   # "awx_plugins.base_interface.api",  # keep `__init__.py` empty
-  "awx_plugins.interfaces",  # standard interface declarations for AWX plugins
+  "awx_plugins.interfaces @git+https://github.com/ansible/awx_plugins.interfaces.git@ad4a965",  # standard interface declarations for AWX plugins
   "PyYAML",  # credentials.injectors, inventory.plugins
   "azure-identity",  # credentials.azure_kv
   "azure-keyvault",  # credentials.azure_kv
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,7 +18,7 @@ dependencies = [  # runtime deps  # https://packaging.python.org/en/latest/guide
   # GUIDANCE: only add things that this project imports directly
   # GUIDANCE: only set lower version bounds
   # "awx_plugins.base_interface.api",  # keep `__init__.py` empty
-  "awx_plugins.interfaces @git+https://github.com/ansible/awx_plugins.interfaces.git@ad4a965",  # standard interface declarations for AWX plugins
+  "awx_plugins.interfaces",  # standard interface declarations for AWX plugins
   "PyYAML",  # credentials.injectors, inventory.plugins
   "azure-identity",  # credentials.azure_kv
   "azure-keyvault",  # credentials.azure_kv
--- a/src/awx_plugins/credentials/aim.py
+++ b/src/awx_plugins/credentials/aim.py
@@ -7,7 +7,7 @@ from awx_plugins.interfaces._temporary_private_django_api import (  # noqa: WPS4
     gettext_noop as _,
 )
 
-import requests
+import requests as requests
 
 from .plugin import CertFiles, CredentialPlugin, raise_for_status
 
--- a/src/awx_plugins/credentials/aim.py
+++ b/src/awx_plugins/credentials/aim.py
@@ -112,14 +112,18 @@ def aim_backend(**kwargs):
             allow_redirects=False,
         )
     sensitive_query_params = {
-        'AppId' : "****",
-        'Query' : "****",
-        'QueryFormat' : object_query_format,   
+        'AppId': '****',
+        'Query': '****',
+        'QueryFormat': object_query_format,
     }
     if reason:
-        sensitive_query_params['reason'] = "****"
-    sensitive_request_qs = '?' + urlencode(sensitive_query_params, safe="*", quote_via=quote)
-    res.url = request_url + sensitive_request_qs
+        sensitive_query_params['reason'] = '****'
+    sensitive_request_qs = urlencode(
+        sensitive_query_params,
+        safe='*',
+        quote_via=quote,
+    )
+    res.url = f'{request_url}?{sensitive_request_qs}'
 
     raise_for_status(res)
     # CCP returns the property name capitalized, username is camel case
--- a/src/awx_plugins/credentials/aim.py
+++ b/src/awx_plugins/credentials/aim.py
@@ -111,6 +111,16 @@ def aim_backend(**kwargs):
             verify=verify,
             allow_redirects=False,
         )
+    sensitive_query_params = {
+        'AppId' : "****",
+        'Query' : "****",
+        'QueryFormat' : object_query_format,   
+    }
+    if reason:
+        sensitive_query_params['reason'] = "****"
+    sensitive_request_qs = '?' + urlencode(sensitive_query_params, safe="*", quote_via=quote)
+    res.url = request_url + sensitive_request_qs
+
     raise_for_status(res)
     # CCP returns the property name capitalized, username is camel case
     # so we need to handle that case
--- a/tests/credential_plugins_test.py
+++ b/tests/credential_plugins_test.py
@@ -6,7 +6,7 @@ from unittest import mock
 import pytest
 
 from awx_plugins.credentials import hashivault
-
+from awx_plugins.credentials import aim
 
 def test_imported_azure_cloud_sdk_vars() -> None:
     from awx_plugins.credentials import azure_kv
@@ -162,3 +162,12 @@ class TestDelineaImports:
         ):
             # assert this module as opposed to older thycotic.secrets.server
             assert cls.__module__ == 'delinea.secrets.server'
+
+def test_aim_sensitive_traceback():
+    aim.aim_backend(
+        url='https://google.com',
+        app_id='test',
+        object_query='test',
+        object_query_format='test',
+        verify=True,
+    )
--- a/tests/credential_plugins_test.py
+++ b/tests/credential_plugins_test.py
@@ -176,8 +176,12 @@ def test_aim_sensitive_traceback_masked(
     my_response.status_code = 404
     my_response.url = 'not_found'
 
-    monkeypatch.setattr(aim.requests, 'get', mocker.Mock(name='aim_request'))
-    aim.requests.get.return_value = my_response
+    aim_request_mock = mocker.Mock(
+        autospec=True,
+        name='aim_request',
+        return_value=my_response,
+    )
+    monkeypatch.setattr(aim.requests, 'get', aim_request_mock)
 
     expected_url_in_exc = (
         r'.*http://testurl\.com/AIMWebService/api/Accounts\?'
@@ -188,7 +192,10 @@ def test_aim_sensitive_traceback_masked(
         'AppId=****&Query=****&QueryFormat=test&reason=****'
     )
 
-    with pytest.raises(requests.exceptions.HTTPError, match=expected_url_in_exc) as e:
+    with pytest.raises(
+        requests.exceptions.HTTPError,
+        match=expected_url_in_exc,
+    ) as e:
         aim.aim_backend(
             url='http://testurl.com',
             app_id='foobar123',
@@ -198,5 +205,5 @@ def test_aim_sensitive_traceback_masked(
             verify=True,
         )
 
-    assert e._excinfo[1].response.url == expected_response_url_literal
+    assert e.value.response.url == expected_response_url_literal
     assert 'foobar123' not in str(e)
--- a/tests/credential_plugins_test.py
+++ b/tests/credential_plugins_test.py
@@ -5,8 +5,11 @@ from unittest import mock
 
 import pytest
 
-from awx_plugins.credentials import hashivault
-from awx_plugins.credentials import aim
+import requests
+from pytest_mock import MockerFixture
+
+from awx_plugins.credentials import aim, hashivault
+
 
 def test_imported_azure_cloud_sdk_vars() -> None:
     from awx_plugins.credentials import azure_kv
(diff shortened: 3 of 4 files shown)


thedoubl3j:
just to note as well @webknjaz it looks like one of the linters that formats code added in a `\` for a line break that then causes https://github.com/wemake-services/flake8-broken-line#error-codes to be thrown. Not sure which one but we might want to update it so we don't have one linter causing another one to error.

pyproject.toml
```diff
@@ -18,7 +18,7 @@ dependencies = [  # runtime deps  # https://packaging.python.org/en/latest/guide
   # GUIDANCE: only add things that this project imports directly
   # GUIDANCE: only set lower version bounds
   # "awx_plugins.base_interface.api",  # keep `__init__.py` empty
-  "awx_plugins.interfaces",  # standard interface declarations for AWX plugins
+  "awx_plugins.interfaces @git+https://github.com/ansible/awx_plugins.interfaces.git@ad4a965",  # standard interface declarations for AWX plugins
```
webknjaz: This piece of metadata must not mention any sources. It's abstract. Put it into the constraints instead: https://github.com/ansible/awx-plugins/blob/9bcd155/dependencies/direct/py-constraints.in#L17.

src/awx_plugins/credentials/aim.py
```diff
@@ -111,6 +111,19 @@ def aim_backend(**kwargs):
...
+        'Query': '****',
+        'QueryFormat': object_query_format,
+    }
+    if reason:
+        sensitive_query_params['reason'] = '****'
+    sensitive_request_qs = (
+        '?' +
+        urlencode(sensitive_query_params, safe='*', quote_via=quote)
+    )
+    res.url = request_url + sensitive_request_qs
```
webknjaz: I'd recommend moving the delimiter to the concatenation instruction. It's also not a part of a query string per RFCs so putting it into a variable called like this would be counterintuitive/misleading:
```suggestion
    sensitive_request_qs = urlencode(
        sensitive_query_params,
        safe='*',
        quote_via=quote,
    )
    res.url = f'{request_url}?{sensitive_request_qs}'
```
P.S. f-strings are always more performant than any other concatenation methods, use them whenever possible.

tests/credential_plugins_test.py
```diff
@@ -162,3 +162,34 @@ def test_tss_import(self) -> None:
...
+                object_query_format='test',
+                reason='foobar123',
+                verify=True,
+            )
+
+        assert """http://testurl.com/AIMWebService/api/Accounts?AppId=
+            ****&Query=****&QueryFormat=test&reason=****""" in str(e)
+        assert e._excinfo[1].response.url == """http://testurl.com/AIMWebService
+            /api/Accounts?AppId=****&Query=****&QueryFormat=test&reason=****"""
+        assert 'foobar123' not in str(e)
```
webknjaz: There are a number of things to improve here:
1) Drop the class namespace — it's just one test, no need to nest them without a good reason. In the future, it'll move into a separate project and wouldn't be surrounded by the unrelated ones anyway.
2) Let's not make a habit of sticking imports inside fixtures. Old tests have that, but we shouldn't follow bad practices going forward.
3) No need to import `unittest` directly, use `pytest-mock` that's integrated and available right away.
4) `pytest.raises()` should always have a `match` regex passed.
5) Awkward formatting can be avoided by putting bigger chunks of data into variables. This also improves readability, as said data can now be labeled by choosing good variable names.
```suggestion
def test_aim_sensitive_traceback_masked(mocker: MockerFixture) -> None:
    """Ensure that the sensitive information is not leaked in the traceback."""
    my_response = requests.Response()
    my_response.status_code = 404
    my_response.url = 'not_found'

    aim.requests.get = mocker.Mock(name='aim_request')
    aim.requests.get.return_value = my_response

    expected_url_in_exc = (
        r'.*http://testurl\.com/AIMWebService/api/Accounts\?'
        r'AppId=\*\*\*\*&Query=\*\*\*\*&QueryFormat=test&reason=\*\*\*\*.*'
    )
    expected_response_url_literal = (
        'http://testurl.com/AIMWebService/api/Accounts?'
        'AppId=****&Query=****&QueryFormat=test&reason=****'
    )

    with pytest.raises(requests.exceptions.HTTPError, match=expected_url_in_exc) as e:
        aim.aim_backend(
            url='http://testurl.com',
            app_id='foobar123',
            object_query='foobar123',
            object_query_format='test',
            reason='foobar123',
            verify=True,
        )

    assert e._excinfo[1].response.url == expected_response_url_literal
    assert 'foobar123' not in str(e)
```

And put the following at the top:
```python
import requests
from pytest_mock import MockerFixture
from awx_plugins.credentials import aim
```

webknjaz (changes requested):
> just to note as well @webknjaz it looks like one of the linters that formats code added in a `\` for a line break that then causes [wemake-services/flake8-broken-line#error-codes](https://github.com/wemake-services/flake8-broken-line#error-codes) to be thrown. Not sure which one but we might want to update it so we don't have one linter causing another one to error.

@thedoubl3j it's an inevitable consequence of the formatter (I'm assuming autopep8 made the change) being separate from the linters. It's also because the formatter uses one of the variants that are compliant with PEP 8 but that the rule disallows that specific variant. Trailing backslashes are dangerous, it's good to disallow them. If you use different compliant formatting, autopep8 won't change it. Perhaps, I'll check out Ruff at some point to see if it could be as non-invasive as autopep8, but I'm not sure.
Additionally, such cases cannot be fully avoided for as long as there's more than one linting/formatting tool integrated and must be addressed by a human.
Another side of this is whenever linting issues arise, they often alert you about the possibility to refactor. In the case you hit, for example, the line was complex. A formatter would make it formatted, but formatted and good are not equal things. These situations always require a human to think about whatever's the underlying issue that tripped the alert.


On another note, find some concrete suggestions below.

pyproject.toml
thedoubl3j: Do we think what is there is good enough already? I think the reason I put this here was because when I was trying to test things, it was screaming about missing something so I dropped it in here.

webknjaz: Well, we might need to do pinning in that constraint file later on, but it shouldn't be in this PR, I think.

Locally, you might've just hit the cache. You can pass `-r` to tox to force it to recreate its venvs.

I've hit caching similar issues with `pre-commit`, which is why I've started pinning the interfaces there. It can be addressed locally but in the `pre-commit.ci` service, such cache is immortal. So yes, we do need to handle this in general, where necessary. If we get to version the libs properly, this will become easier.

thedoubl3j: ahh gotcha, ty for the `-r`

webknjaz: @thedoubl3j Note that `-r` will only clean tox's own venvs. If you want to clean what the `pre-commit` tool manages, you could try `tox e -e pre-commit -- pytest -Im pre_commit clean` and it'll clear the userspace venvs it manages.

tests/credential_plugins_test.py
```diff
@@ -162,3 +165,35 @@ def test_tss_import(self) -> None:
...
             assert cls.__module__ == 'delinea.secrets.server'
+
+
+def test_aim_sensitive_traceback_masked(mocker: MockerFixture) -> None:
+    """Ensure that the sensitive information is not leaked in the traceback."""
+    my_response = requests.Response()
+    my_response.status_code = 404
+    my_response.url = 'not_found'
+
+    aim.requests.get = mocker.Mock(name='aim_request')
```
webknjaz: I didn't notice this before, but this is mutating the process-global state and doesn't restore it after. To fix this, either use `monkeypatch.setattr()` or `mocker.patch` both of which are able to undo the changes..

webknjaz:
```suggestion
def test_aim_sensitive_traceback_masked(mocker: MockerFixture, monkeypatch: pytest.MonkeyPatch) -> None:
    """Ensure that the sensitive information is not leaked in the traceback."""
    my_response = requests.Response()
    my_response.status_code = 404
    my_response.url = 'not_found'

    monkeypatch.setattr(aim.requests, 'get', mocker.Mock(name='aim_request'))
```

pyproject.toml
webknjaz:
```suggestion
  "awx_plugins.interfaces",  # standard interface declarations for AWX plugins
```

webknjaz:
pre-commit.ci run

tests/credential_plugins_test.py
```diff
@@ -162,3 +165,44 @@ def test_tss_import(self) -> None:
...
+def test_aim_sensitive_traceback_masked(
+        mocker: MockerFixture,
+        monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    """Ensure that the sensitive information is not leaked in the traceback."""
+    my_response = requests.Response()
+    my_response.status_code = 404
+    my_response.url = 'not_found'
+
+    mock_get = mocker.patch(
```
webknjaz: why didn't you keep monkeypatch?

.pre-commit-config.yaml
webknjaz, line 1: You missed the pylint one. But don't worry, I'm going to address that on `devel` directly.

webknjaz: Plz rebase now.

tests/credential_plugins_test.py
thedoubl3j: not sure, lemme go see. I wonder if something happened in a conflict merge.

webknjaz: Probably, since you didn't drop the fixture I injected from the function args, just stopped using it.

.pre-commit-config.yaml
```diff
@@ -201,6 +201,7 @@ repos:
     - lxml  # dep of `--txt-report`, `--cobertura-xml-report` & `--html-report`
     - msrestazure  # needed by credentials.azure_kv
     - pytest
+    - pytest_mock
```
webknjaz: This will disappear post-rebase. The PR is blocked on conflict resolution: https://results.pre-commit.ci/run/github/836873755/1729693574.gKfyS5-rR2mqX3tQ-7MlzA.

tests/credential_plugins_test.py
webknjaz: Marking this as resolved.

tests/credential_plugins_test.py
webknjaz: Marking this as resolved.

.pre-commit-config.yaml
webknjaz: marking as resolved

.pre-commit-config.yaml
webknjaz: marking as resolved.

tests/credential_plugins_test.py
```diff
@@ -162,3 +165,41 @@ def test_tss_import(self) -> None:
...
+        mocker: MockerFixture,
+        monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    """Ensure that the sensitive information is not leaked in the traceback."""
+    my_response = requests.Response()
+    my_response.status_code = 404
+    my_response.url = 'not_found'
+
+    monkeypatch.setattr(aim.requests, 'get', mocker.Mock(name='aim_request'))
+    aim.requests.get.return_value = my_response
```
webknjaz: try this
```suggestion
    aim_request_mock = mocker.Mock(
        autospec=True,
        name='aim_request',
        return_value = my_response,
    )
    monkeypatch.setattr(aim.requests, 'get', aim_request_mock)
```

webknjaz (approved):

djyasin (approved):

chrismeyersfsu (approved):

tests/credential_plugins_test.py
```diff
@@ -162,3 +165,89 @@ def test_tss_import(self) -> None:
...
+    ],
+)
+def test_aim_sensitive_traceback_masked(
+    mocker: MockerFixture,
+    monkeypatch: pytest.MonkeyPatch,
+    url: str,
+    app_id: str,
+    object_query: str,
+    object_query_format: str,
+    reason: str | None,
```
webknjaz: Looking into the params, I see that it's always `str`. There's no test that checks `None`.

```suggestion
    reason: str,
```

Or did you want to add another param with `None` additionally to an empty string?

tests/credential_plugins_test.py
```diff
@@ -162,3 +165,89 @@ def test_tss_import(self) -> None:
...
+
+
+@pytest.mark.parametrize(
+    (
+        'url',
+        'app_id',
+        'object_query',
+        'object_query_format',
+        'reason',
+        'verify',
```
webknjaz: Out of all these, only `reason` differs. So I don't see why they should be duplicated and declared as params.

Parametrizing just `reason` would be equivalent and more than enough.

Although, looking into the rest of the commit, I see that it makes sense to add two other params. But only for things that actually differ.

tests/credential_plugins_test.py
```diff
@@ -162,3 +165,89 @@ def test_tss_import(self) -> None:
...
+        )
+    else:
+        expected_url_in_exc = (
+            r'.*http://testurl\.com/AIMWebService/api/Accounts\?'
+            r'AppId=\*\*\*\*&Query=\*\*\*\*&QueryFormat=test.*'
+        )
+        expected_response_url_literal = (
+            'http://testurl.com/AIMWebService/api/Accounts?'
+            'AppId=****&Query=****&QueryFormat=test'
+        )
```
webknjaz: It's best to avoid any logic in the tests. They should be as straightforward as possible. Doing otherwise complicates things unnecessarily and is prone to human error and accumulating complexity.

The solution is simple — parametrize these two vars just like `reason` so each parametrized test would have 3 inputs — reason and two expectations. You'd just pass them to the thing you're testing and check the assertion. No extra logic.

webknjaz (dismissed):
Looks like there's a few things to correct.

tests/credential_plugins_test.py
thedoubl3j: I think just an empty str actually covers it. change reverted

tests/credential_plugins_test.py
```diff
@@ -162,3 +165,62 @@ def test_tss_import(self) -> None:
...
+     'expected_url_in_exc',
+     'expected_response_url_literal',
+     ),
+    (pytest.param(
+        'foobar123',
+        r'.*http://testurl\.com/AIMWebService/api/Accounts\?'
+        r'AppId=\*\*\*\*&Query=\*\*\*\*&QueryFormat=test&reason=\*\*\*\*.*',
+        'http://testurl.com/AIMWebService/api/Accounts?'
+        'AppId=****&Query=****&QueryFormat=test&reason=****',
+        id='with_reason',
```
webknjaz:
```suggestion
        id='with-reason',
```

tests/credential_plugins_test.py
```diff
@@ -162,3 +165,62 @@ def test_tss_import(self) -> None:
...
+    ),
+        pytest.param(
+            '',
+            r'.*http://testurl\.com/AIMWebService/api/Accounts\?'
+            r'AppId=\*\*\*\*&Query=\*\*\*\*&QueryFormat=test.*',
+            'http://testurl.com/AIMWebService/api/Accounts?'
+            'AppId=****&Query=****&QueryFormat=test',
+            id='no_reason',
```
webknjaz:
```suggestion
            id='no-reason',
```

webknjaz:
pre-commit.ci run

tests/credential_plugins_test.py
```diff
@@ -162,3 +165,69 @@ def test_tss_import(self) -> None:
...
+            (
+                r'.*http://testurl\.com/AIMWebService/api/Accounts\?'
+                r'AppId=\*\*\*\*&Query=\*\*\*\*&QueryFormat=test&'
+                r'reason=\*\*\*\*.*'
+            ),
+            (
+                'http://testurl.com/AIMWebService/api/Accounts?'
+                'AppId=****&Query=****&QueryFormat=test&reason=****'
+            ),
+            id='with_reason',
```
webknjaz:
```suggestion
            id='with-reason',
```

webknjaz (approved):

webknjaz (approved):

PR #52, opened 2024-10-29
Split the unconditional runtime dependencies into extras

jessicamack:
All the runtime dependencies of `awx-plugins-core` are unconditional now. To take into account being able to depend only on certain plugins, I've split them into optional ones.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,15 +18,6 @@ dependencies = [  # runtime deps  # https://packaging.python.org/en/latest/guide
   # GUIDANCE: only add things that this project imports directly
   # GUIDANCE: only set lower version bounds
   # "awx_plugins.base_interface.api",  # keep `__init__.py` empty
-  "awx_plugins.interfaces",  # standard interface declarations for AWX plugins
-  "PyYAML",  # credentials.injectors, inventory.plugins
-  "azure-identity",  # credentials.azure_kv
-  "azure-keyvault",  # credentials.azure_kv
-  "boto3",  # credentials.awx_secretsmanager
-  "msrestazure",  # credentials.azure_kv
-  "python-dsv-sdk >= 1.0.4",  # credentials.thycotic_dsv
-  "python-tss-sdk >= 1.2.1",  # credentials.thycotic_tss
-  "requests",  # credentials.aim, credentials.centrify_vault, credentials.conjur, credentials.hashivault
 ]
 classifiers = [  # Allowlist: https://pypi.org/classifiers/
   "Development Status :: 1 - Planning",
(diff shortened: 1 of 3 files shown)


webknjaz:
@jessicamack to fix the CI, change `pyproject.toml` to `pyproject.toml --all-extras` on these lines: https://github.com/ansible/awx-plugins/blob/c7fc0a1/.github/workflows/pip-tools.yml#L102-L128. Also, here https://github.com/ansible/awx-plugins/blob/c7fc0a1/tox.ini#L15, after `deps =`, add `extras =` with a column containing all the extras, one per line, use two-space indents.

pyproject.toml
```diff
@@ -14,20 +14,11 @@ name = "awx-plugins-core"  # import awx_plugins.credentials.x, awx_plugins.crede
...
-  "PyYAML",  # credentials.injectors, inventory.plugins
-  "azure-identity",  # credentials.azure_kv
-  "azure-keyvault",  # credentials.azure_kv
-  "boto3",  # credentials.awx_secretsmanager
-  "msrestazure",  # credentials.azure_kv
-  "python-dsv-sdk >= 1.0.4",  # credentials.thycotic_dsv
-  "python-tss-sdk >= 1.2.1",  # credentials.thycotic_tss
-  "requests",  # credentials.aim, credentials.centrify_vault, credentials.conjur, credentials.hashivault
-]
+# dependencies = [  # runtime deps  # https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#dependencies-optional-dependencies
```
webknjaz: Keep the empty list declared instead of commenting this out.

webknjaz:
@jessicamack have you seen https://github.com/ansible/awx-plugins/pull/52#issuecomment-2445712910?

djyasin (approved):

webknjaz:
UPD: this is failing because of a bug in tox (https://github.com/tox-dev/tox/issues/3433) which is blocking the PR.

pyproject.toml
```diff
@@ -105,14 +96,97 @@ file = "LICENSE"
...
+[project.optional-dependencies]
+# These are additional runtime dependencies used as feature-flags.
+# The end-users will be able to reference them in their
+# dependencies. Do not abuse as "test" / "dev" deps.
+# Treat as public API.
+credentials-conjur = [
+  "awx_plugins.interfaces",
+  "requests",
+]
+credentials-hashivault_kv = [
```
webknjaz: Example:
```suggestion
credentials-hashivault-kv = [
```

tox.ini
```diff
@@ -13,6 +13,28 @@ warnings-to-errors = -Werror
 [testenv]
 description = Run pytest under {envpython}
 deps = -rdependencies{/}direct{/}py.in
+extras =
+  credentials-conjur
+  credentials-hashivault_kv
```
webknjaz: Example:
```suggestion
  credentials-hashivault-kv
```

tox.ini
```diff
@@ -555,6 +581,7 @@ commands =
 commands_post =
 deps =
   -rdependencies{/}direct{/}{envname}.in
+extras = {[testenv:build-docs]extras}
```
webknjaz: revert
```suggestion
```

tox.ini
```diff
@@ -519,6 +544,7 @@ commands_post =
       gh_summary_fd.close()'
 deps =
   -rdependencies{/}direct{/}{envname}.in
+extras = {[testenv:build-docs]extras}
```
webknjaz: revert
```suggestion
```

tox.ini
```diff
@@ -466,6 +490,7 @@ commands_post =
       gh_summary_fd.close()'
 deps =
   -r{toxinidir}{/}dependencies{/}direct{/}{envname}.in
+extras = {[testenv:build-docs]extras}
```
webknjaz: revert
```suggestion
```

tox.ini
```diff
@@ -416,6 +439,7 @@ commands_post =
...
```
webknjaz: revert
```suggestion
```

tox.ini
```diff
@@ -365,6 +387,7 @@ commands_post =
       \N\{QUOTATION MARK\}\{docs_dir\}\N\{QUOTATION MARK\} 0\n")'
 deps =
   -r{toxinidir}{/}dependencies{/}direct{/}{envname}.in
+extras = {[testenv]extras}
```
webknjaz: revert
```suggestion
```

webknjaz (dismissed):
@jessicamack so I learned that [PEP 685](https://packaging.python.org/en/latest/specifications/core-metadata/#provides-extra-multiple-use) will make the format of extras stricter in the internal core packaging metadata version 2.3. The user-facing extra names might still be allowed to be non-normalized. But in general, it's best to just follow that format to avoid confusion. Evidently, some tools are going to get confused and hit corner cases if we don't do this.

Action items:
* [x] Change extra names in `pyproject.toml` to have `-` instead of `_`
* [x] Change extra names in `tox.ini` to have `-` instead of `_`

webknjaz:
@chrismeyersfsu could you confirm that the extra names look fine? Not that they can't be changed, but it's a chance to update them if there's anything obvious that stands out.

webknjaz (approved):
Waiting from Chris' take on the name, and then we can merge.

chrismeyersfsu:
So I understand, after this merges, awx will NOT get these extra modules from awx-plugins.interfaces unless awx adds the optional directive I.e. awx-plugins[x,y,z]

webknjaz:
Yes. Although, it could just start depending on the version you released on Monday until things settle and they those extras that remain post split activities would need to be listed along with the version bump.

chrismeyersfsu (approved):

webknjaz:
Now a companion dep update is needed in awx.

PR #55, opened 2024-11-10
Update aap credential injectors

sean-m-sullivan:
Update the injectors for the AAP credential type to work across collections

Related PR's:

- [awx-plugins](https://github.com/ansible/awx-plugins/pull/55)
- [ansible.platform](https://github.com/ansible/aap-gateway/pull/627)
- [ansible.hub](https://github.com/ansible-collections/ansible_hub/pull/29)
- [ansible.controller](https://github.com/ansible/awx/pull/15627)
- [ansible.eda](https://github.com/ansible/event-driven-ansible/pull/347)
- [eda-server](https://github.com/ansible/eda-server/pull/1162)
- infra.aap_configuration - Done.
- https://issues.redhat.com/browse/ANSTRAT-1181 consolidates this

### DIFF
--- a/src/awx_plugins/credentials/plugins.py
+++ b/src/awx_plugins/credentials/plugins.py
@@ -590,8 +590,7 @@ insights = ManagedCredentialType(
                 'label': gettext_noop('Username'),
                 'type': 'string',
                 'help_text': gettext_noop(
-                    'Required for basic authentication. '
-                    'May be blank if using client_id and client_secret',
+                    'Username is required for basic authentication.',
                 ),
             },
             {
@@ -600,8 +599,7 @@ insights = ManagedCredentialType(
                 'type': 'string',
                 'secret': True,
                 'help_text': gettext_noop(
-                    'Required for basic authentication. '
-                    'May be blank if using client_id and client_secret',
+                    'Password is required for basic authentication.',
                 ),
             },
             {
@@ -609,8 +607,7 @@ insights = ManagedCredentialType(
                 'label': gettext_noop('Client ID'),
                 'type': 'string',
                 'help_text': gettext_noop(
-                    'Required for service account authentication. '
-                    'May be blank if using username and password',
+                    'Enter client ID to create a service account credential.',
                 ),
             },
             {
@@ -619,8 +616,7 @@ insights = ManagedCredentialType(
                 'type': 'string',
                 'secret': True,
                 'help_text': gettext_noop(
-                    'Required for service account authentication. '
-                    'May be blank if using username and password',
+                    'Enter client secret to create a service account credential.',
                 ),
             },
         ],


willtome:
I ran into the issue as well. Very frustrating and needs to be fixed

src/awx_plugins/credentials/plugins.py
```diff
@@ -688,10 +688,30 @@
...
     },
     injectors={
+        'extra_vars': {
+            'aap_hostname': '{{host}}',
+            'aap_username': '{{password}}',
+            'aap_password': '{{password}}',
+            'aap_token': '{{oauth_token}}',
+            'aap_request_timeout': '{{request_timeout}}',
+            'aap_validate_certs': '{{verify_ssl}}',
+            'ah_path_prefix': '{{ah_path_prefix}}',
```
AlanCoding: extra_vars never did anything. That's why they were never put in the injectors before. These appears to correspond to the module arguments, but that implies a syntax like `aap_hostname: "{{ aah_hostname }}"` used in your playbook. This is far from obvious, and I'm not convinced it's useful. It would be very undocumented. Very hard to document.

src/awx_plugins/credentials/plugins.py
```diff
@@ -688,10 +688,30 @@
...
             },
+            {
+                'id': 'ah_path_prefix',
+                'label': gettext_noop('Verify SSL'),
+                'type': 'string',
+                'secret': False,
+                'default': 'galaxy',
+                'help_text': gettext_noop(
+                    'API path used to access the Hub api. Either galaxy, '
+                    'or custom. By default it should be galaxy',
```
AlanCoding: > Either galaxy or custom

I've never seen this. I can't find it anywhere in AWX. I would ask Hub engineers what the intent is... but this is the "controller" credential type. It doesn't add up to me.

src/awx_plugins/credentials/plugins.py
```diff
@@ -703,6 +723,14 @@
             'CONTROLLER_PASSWORD': '{{password}}',
             'CONTROLLER_VERIFY_SSL': '{{verify_ssl}}',
             'CONTROLLER_OAUTH_TOKEN': '{{oauth_token}}',
+            'CONTROLLER_REQUEST_TIMEOUT': '{{request_timeout}}',
+            'AAP_HOSTNAME': '{{host}}',
+            'AAP_USERNAME': '{{username}}',
+            'AAP_PASSWORD': '{{password}}',
+            'AAP_VALIDATE_CERTS': '{{verify_ssl}}',
+            'AAP_TOKEN': '{{oauth_token}}',
+            'AAP_REQUEST_TIMEOUT': '{{request_timeout}}',
```
AlanCoding: These seem fine up to this point.

willtome: will TOWER_HOST, CONTROLLER_HOST, and AAP_HOSTNAME ever need to be different?

AlanCoding: Yes, that is the point. Using CONTROLLER_HOST by itself is using direct API access to the controller host, meaning, you are not going through the proxy. That way of accessing controller is deprecated (exactly the meaning of that word, I'm not the one to ask), so we want to be moving away from that.

However, this change is allowing to use the address of the proxy host. I think @chrismeyersfsu might have expressed a sentiment that this should be a separate type, and there is some validity to that. Ultimately, if we want existing credentials to "just work", we may need some code-level solution to this so that it inspects the response headers or something to determine which it was given, because this is expected to affect the API path that has to be used.

sean-m-sullivan: @willtome You can point controller_host at the gateway and things work, I tested this all against a gateway 2.5 address like a customer should be doing, The variable itself doesn't care about proxy, endpoint, etc etc. From what I've tested awx.awx against a 2.5 AAP will not work as it does not do the api translation correctly, however ansible.controller collection does. Technically with upstream it could just be Device FQDN, but we have to call it something, and the user relies on the collection to know the difference and know better.

src/awx_plugins/credentials/plugins.py
sean-m-sullivan: We were writing a blog post on it, but if we just document they are avail, or make them avail, people can use them in places to connect, or use for the lokup plugin or other uses. We shouldn't limite HOW they are used just make them available for the base credential type. I and probably everyone using the validated for a while have also been using custom, just trying to take the standard custom and bring it into the fold.

src/awx_plugins/credentials/plugins.py
sean-m-sullivan: It's https://10.242.42.153/api/galaxy/

I just checked, in 2.4 it was customizable in the installer through an obscure variable, with 2.5 it looks like it was hardcoded. 

I've removed this entry.

src/awx_plugins/credentials/plugins.py
```diff
@@ -744,10 +744,29 @@
...
     },
     injectors={
+        'extra_vars': {
+            'aap_hostname': '{{host}}',
+            'aap_username': '{{username}}',
+            'aap_password': '{{password}}',
+            'aap_token': '{{oauth_token}}',
+            'aap_request_timeout': '{{request_timeout}}',
+            'aap_validate_certs': '{{verify_ssl}}',
+        },
```
AlanCoding: I'm only slightly worried that adding extra_vars may introduce compatibility issues, because some credential uses don't take extra_vars, like inventory imports. But I haven't been involved much with awx-plugins and I'd like someone who has a little more recent experience to comment.

webknjaz:
FTR, the broken link in docs has been fixed on `devel`. Rebasing this PR should clear the job failure.

src/awx_plugins/credentials/plugins.py
sean-m-sullivan: @AlanCoding I don't think this will be picked up for inventories, and its not built to do that, This was added specifically to work with the aap_configuration collection, and if the user was wanting to use it in a playbook, Have no clue if it would work with inventories, how does it work now(I dont think it works as is with extra vars, or at least defaults to the env vars)

AlanCoding: https://github.com/ansible/awx/blob/devel/awx/main/tasks/jobs.py#L531

https://github.com/ansible/awx/blob/devel/awx/main/models/credential.py#L553

https://github.com/ansible/awx_plugins.interfaces/blob/9af86ce7543377c63dabeb373159abc37bfa8ff4/src/awx_plugins/interfaces/_temporary_private_inject_api.py#L155

You see way down at the bottom it still seems to do:

```python
    if 'INVENTORY_UPDATE_ID' not in env:
        extra_vars = _build_extra_vars(
            sandbox_env,
            namespace,
            cred_type.injectors.get(
                'extra_vars', {},
            ),
        )
```

which is kind of terrible and klunky. But I guess it is at least expected to work.

AlanCoding (approved):

sean-m-sullivan:
@AlanCoding

pb82:
@sean-m-sullivan sorry, I think I messed up your branch when I wanted to rebase it :( I'll create a new PR with your commits.

webknjaz:
@pb82 this is also a case of https://hynek.me/articles/pull-requests-branch/

PR #56, opened 2024-11-14
Add descriptions for plugins

djyasin:
Add descriptions for plugins

### DIFF
--- a/src/awx_plugins/inventory/plugins.py
+++ b/src/awx_plugins/inventory/plugins.py
@@ -18,6 +18,7 @@ class PluginFileInjector:
     plugin_name = None  # Ansible core name used to reference plugin
     # base injector should be one of None, "managed", or "template"
     # this dictates which logic to borrow from playbook injectors
+    plugin_description = None  # this field holds the injector description
     base_injector = None
     # every source should have collection, these are for the collection name
     namespace = None
@@ -150,6 +151,7 @@ class PluginFileInjector:
 
 class azure_rm(PluginFileInjector):
     plugin_name = 'azure_rm'
+    plugin_description = 'Microsoft Azure Resource Manager'
     base_injector = 'managed'
     namespace = 'azure'
     collection = 'azcollection'
@@ -163,6 +165,7 @@ class azure_rm(PluginFileInjector):
 
 class ec2(PluginFileInjector):
     plugin_name = 'aws_ec2'
+    plugin_description = 'Amazon EC2'
     base_injector = 'managed'
     namespace = 'amazon'
     collection = 'aws'
@@ -176,6 +179,7 @@ class ec2(PluginFileInjector):
 
 class gce(PluginFileInjector):
     plugin_name = 'gcp_compute'
+    plugin_description = 'Google Compute Engine'
     base_injector = 'managed'
     namespace = 'google'
     collection = 'cloud'
@@ -198,6 +202,7 @@ class gce(PluginFileInjector):
 
 class vmware(PluginFileInjector):
     plugin_name = 'vmware_vm_inventory'
+    plugin_description = 'VMware vCenter'
     base_injector = 'managed'
     namespace = 'community'
     collection = 'vmware'
@@ -205,6 +210,7 @@ class vmware(PluginFileInjector):
 
 class openstack(PluginFileInjector):
     plugin_name = 'openstack'
+    plugin_description = 'OpenStack'
     namespace = 'openstack'
     collection = 'cloud'
 
@@ -264,6 +270,7 @@ class rhv(PluginFileInjector):
     """Ovirt uses the custom credential templating, and that is all."""
 
     plugin_name = 'ovirt'
+    plugin_description = 'Red Hat Virtualization'
     base_injector = 'template'
     initial_version = '2.9'
     namespace = 'ovirt'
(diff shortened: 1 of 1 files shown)


src/awx_plugins/inventory/plugins.py
```diff
@@ -1,5 +1,5 @@
 # FIXME: the following violations must be addressed gradually and unignored
-# mypy: disable-error-code="assignment, no-untyped-call, no-untyped-def, var-annotated"
+# mypy: disable-error-code='assignment, no-untyped-call, no-untyped-def, var-annotated'
```
webknjaz: I'm assuming a misconfigured code editor made a bunch of invasive changes, eh?

djyasin: I ran black on this before I committed and it reformatted everything- I am going to pull out those changes today!

src/awx_plugins/inventory/plugins.py
```diff
@@ -18,6 +18,8 @@ class PluginFileInjector:
     plugin_name = None  # Ansible core name used to reference plugin
     # base injector should be one of None, "managed", or "template"
     # this dictates which logic to borrow from playbook injectors
+    plugin_description = None
+    # this field is for the description of the injector
```
webknjaz: In general, comments should be on the same line if possible:
```suggestion
    plugin_description = None  # this field holds the injector description
```
This is because over time, the comment line and the code line may drift away from each other by accident. And the comments will end up being in weird places.

webknjaz (dismissed):
Does this still need to remain a draft?

P.S. Note that all of the revert+fixup+wip commit titles will end up on `devel`.

src/awx_plugins/inventory/plugins.py
webknjaz: Why didn't you run the automation that is integrated into the repo instead?

djyasin: It was just force of habit, I will absolutely do that moving forward!

webknjaz: Totally understandable :)
I was just configuring my editor for the new laptop the other day and the first thing I did was figuring out how to make it non-invasive because the framework I'm on happens to have formatting on save which is a harmful assumption to have universally in context of arbitrary projects I might want to open..

src/awx_plugins/inventory/plugins.py
```diff
@@ -64,6 +69,9 @@ def inventory_as_dict(self, inventory_update, private_data_dir):
             }.{
                 self.plugin_name
             }""" if self.use_fqcn else self.plugin_name
+            {
+                self.plugin_description,
+            }
```
webknjaz:
```suggestion
```

src/awx_plugins/inventory/plugins.py
```diff
@@ -30,7 +31,11 @@ class PluginFileInjector:
     def get_proper_name(cls) -> str | None:
```
webknjaz: If this is used in contexts that expect an FQCN, it'll break things externally.

djyasin: Thank you!

webknjaz: Why did you mark it as resolved, though?

src/awx_plugins/inventory/plugins.py
```diff
@@ -30,7 +31,15 @@ class PluginFileInjector:
...
         if cls.plugin_name is None:
             return None
-        return f'{cls.namespace}.{cls.collection}.{cls.plugin_name}'
+        return f'{
```
webknjaz: If you still want this change for some reason, given that it's likely breaking, it's necessary to fix the syntax and make it a multiline string. Otherwise, revert this change to the working state.

src/awx_plugins/inventory/plugins.py
djyasin: My mistake- I am so sorry! Should be fixed now :)

src/awx_plugins/inventory/plugins.py
djyasin: I have revereted this to its working state. Thank you again!

webknjaz:
@djyasin with all the changes back and forth, Codecov seems to be confused. Could you rebase to see if that'll fix the problem?

djyasin:
> @djyasin with all the changes back and forth, Codecov seems to be confused. Could you rebase to see if that'll fix the problem?

I just  rebased and squashed, let me know if that looks better!

webknjaz:
@djyasin perfect! Codecov no longer reports 0% patch coverage. The commit description could be cleaned up, but that's not a hard blocker. So let's not delay merging any further.

webknjaz (approved):

webknjaz:
@djyasin *post*-merge instructions:
1. Click `Run workflow` @ https://github.com/ansible/awx-plugins/actions/workflows/ci-cd.yml, entering `0.0.1a7` precisely in the version field. Don't enter anything else, and submit the form.
2. Wait until the newly created workflow run completes tests and pauses, then approve releasing on the workflow run summary page.
3. Wait until the workflow run completes fully.
4. Make sure awx gets the new version. Make a PR bumping the dependency over there if necessary.

PR #50, opened 2024-10-22
📦 Discover injectors through entrypoints

webknjaz:
This patch eliminates import-based injector discovery, switching to
distribution package-independent hosting of the respective callables.

PR #57, opened 2024-11-15
Move injectors to ManagedCredentialType

chrismeyersfsu:
Injectors are a bail-out mechanism for when you need to do something that the templating engine does not support, for a particular credential type. They are strongly tied to one and only one credential type. Before this change they lived far away from the ManagedCredentialType they are associated for. The link was the credential kind string was the same as the injector function name. This was a very loose coupling. This change is a tighter coupling.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -209,4 +209,8 @@ nitpicky = True
 # Ref: https://stackoverflow.com/a/30624034/595220
 nitpick_ignore = [
     # temporarily listed ('role', 'reference') pairs that Sphinx cannot resolve
+    (
+        'py:class',
+        'awx_plugins.interfaces._temporary_private_credential_api.Credential',
+    ),
 ]
--- a/src/awx_plugins/credentials/injectors.py
+++ b/src/awx_plugins/credentials/injectors.py
@@ -9,6 +9,10 @@ import tempfile
 from awx_plugins.interfaces._temporary_private_container_api import (  # noqa: WPS436
     get_incontainer_path,
 )
+from awx_plugins.interfaces._temporary_private_credential_api import (  # noqa: WPS436
+    Credential,
+    GenericOptionalPrimitiveType,
+)
 from awx_plugins.interfaces._temporary_private_django_api import (  # noqa: WPS436
     get_vmware_certificate_validation_setting,
 )
@@ -16,7 +20,11 @@ from awx_plugins.interfaces._temporary_private_django_api import (  # noqa: WPS4
 import yaml
 
 
-def aws(cred, env, private_data_dir):
+def aws(
+    cred: Credential,
+    env: dict[str, GenericOptionalPrimitiveType],
+    private_data_dir: str,
+) -> None:
     env['AWS_ACCESS_KEY_ID'] = cred.get_input('username', default='')
     env['AWS_SECRET_ACCESS_KEY'] = cred.get_input('password', default='')
 
@@ -27,7 +35,11 @@ def aws(cred, env, private_data_dir):
         env['AWS_SESSION_TOKEN'] = env['AWS_SECURITY_TOKEN']
 
 
-def gce(cred, env, private_data_dir):
+def gce(
+    cred: Credential,
+    env: dict[str, GenericOptionalPrimitiveType],
+    private_data_dir: str,
+) -> str:
     project = cred.get_input('project', default='')
     username = cred.get_input('username', default='')
 
@@ -66,13 +78,17 @@ def gce(cred, env, private_data_dir):
     return path
 
 
-def azure_rm(cred, env, private_data_dir):
+def azure_rm(
+    cred: Credential,
+    env: dict[str, GenericOptionalPrimitiveType],
+    private_data_dir: str,
+) -> None:
     client = cred.get_input('client', default='')
     tenant = cred.get_input('tenant', default='')
 
     env['AZURE_SUBSCRIPTION_ID'] = cred.get_input('subscription', default='')
 
-    if len(client) and len(tenant):
+    if client and tenant:
         env['AZURE_CLIENT_ID'] = client
         env['AZURE_TENANT'] = tenant
         env['AZURE_SECRET'] = cred.get_input('secret', default='')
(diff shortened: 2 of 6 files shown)


AlanCoding (approved):
Putting this in my own words...

The custom logic like `awx_plugins.credentials.injectors.aws` already lived here, but there was no link in the plugin definition to the custom logic. So somehow that got referenced by a literal name like `"aws"`.

My only gripe would be that `post_injectors` doesn't feel like a _great_ name. Like, sure, maybe it runs after the other stuff (templates)? But it's usually an _alternative_ to the templates, making that detail kind of irrelevant practically. It's more important that it runs custom python logic like a caveman, as opposed to any structured injector definition.

chrismeyersfsu:
1. `injectors_override`
2. `custom_injectors`
3. `alt_injectors`
4. `injectors_fn`

AlanCoding:
I'd lean towards (2) myself.

webknjaz:
I had an alternative idea with https://github.com/ansible/awx_plugins.interfaces/pull/10 / https://github.com/ansible/awx-plugins/pull/50 / https://github.com/ansible/awx/pull/15595 FTR

src/awx_plugins/credentials/plugins.py
```diff
@@ -711,6 +726,7 @@
     namespace='kubernetes_bearer_token',
     kind='kubernetes',
     name=gettext_noop('OpenShift or Kubernetes API Bearer Token'),
+    injector=kubernetes_bearer_token,
```
AlanCoding:
```suggestion
    custom_injector=kubernetes_bearer_token,
```

AlanCoding:
I slightly lean towards the sentiment @chrismeyersfsu expressed in https://github.com/ansible/awx/pull/15595

> So I lean more towards getting rid of this "problem" by bringing these custom injectors closer to the ManagedCredentialType they are related to.

But even if we did nothing, it wouldn't bother me that much, since they already live in the plugins repo.

webknjaz:
@chrismeyersfsu typing change handling instructions:
1. ~Merge https://github.com/ansible/awx_plugins.interfaces/pull/13~
2. ~Click `Run workflow` @ https://github.com/ansible/awx_plugins.interfaces/actions/workflows/ci-cd.yml, entering `0.0.1a2` precisely in the version field. Don't enter anything else and submit the form.~
3. ~Wait until the newly created workflow run completes tests and pauses, then approve releasing on the workflow run summary page.~
4. ~Wait until the workflow run completes fully.~
5. Click `Run workflow` @ https://github.com/ansible/awx-plugins/actions/workflows/pip-tools.yml, entering `awx_plugins.interfaces` in the only input field.
6. Wait until that workflow completes and opens a PR with the version bump in the lock files.
7. Go to that PR and undraft it. You can approve it and enable auto-merge at this point.
8. Once that PR is merged, you can rebase this one.

src/awx_plugins/credentials/injectors.py
```diff
@@ -6,6 +6,9 @@
 import stat
 import tempfile
 
+from awx_plugins.interfaces._temporary_private_api import (  # noqa: WPS436
```
webknjaz:
```suggestion
from awx_plugins.interfaces._temporary_private_credential_api import (  # noqa: WPS436
```

src/awx_plugins/credentials/injectors.py
```diff
@@ -16,7 +19,7 @@
 import yaml
 
 
-def aws(cred, env, private_data_dir):
+def aws(cred: Credential, env: dict, private_data_dir: str):
```
webknjaz: While on it, could you also mark the returned type for all of the functions in the module?
```suggestion
def aws(cred: Credential, env: dict, private_data_dir: str) -> None:
```

src/awx_plugins/credentials/plugins.py
```diff
@@ -8,6 +8,16 @@
     gettext_noop,
 )
 
+from awx_plugins.credentials.injectors import (
```
webknjaz: Always use relative imports for local modules:
```suggestion
from .injectors import (
```

AlanCoding (approved):

src/awx_plugins/credentials/injectors.py
```diff
@@ -9,14 +9,22 @@
...
 
 import yaml
 
 
-def aws(cred, env, private_data_dir):
+def aws(
+    cred: Credential,
+    env: dict[str, GenericOptionalPrimitiveType],
+    private_data_dir: str,
+) -> None:
```
webknjaz, line 27: FWIW, I recommend reverting the hack in #64 and adding docstrings to all of these instead.

.pre-commit-config.yaml
```diff
@@ -194,7 +194,7 @@ repos:
     additional_dependencies:
     - >-  # standard interface declarations for AWX plugins
       awx_plugins.interfaces
-      @ git+https://github.com/ansible/awx_plugins.interfaces.git@ad4a965
+      @ git+https://github.com/ansible/awx_plugins.interfaces.git@ca7802a
```
webknjaz: I mentioned it somewhere already — best to drop Git references, as they are last resort.

```suggestion
    - awx_plugins.interfaces == 0.0.1a2
```

webknjaz: fixed via #65

docs/conf.py
```diff
@@ -209,4 +209,8 @@
 # Ref: https://stackoverflow.com/a/30624034/595220
 nitpick_ignore = [
     # temporarily listed ('role', 'reference') pairs that Sphinx cannot resolve
+    (
+        'py:class',
+        'awx_plugins.interfaces._temporary_private_credential_api.Credential',
+    ),
```
webknjaz, line 215: Recommended workaround: https://github.com/ansible/awx-plugins/pull/24/files#diff-85933aa74a2d66c3e4dcdf7a9ad8397f5a7971080d34ef1108296a7c6b69e7e3R209-R245

TheRealHaoLiu (approved):

AlanCoding (approved):

PR #62, opened 2024-12-02
Work around docutils v0.21.2 deprecation issue

chrismeyersfsu:
* A deprecation warning in docutils surfaces through sphinx-autodoc-typehints. A workaround was submitted to sphinx-autodoc-typehints but that workaround/fix is not yet in a release version. More background info can be found here
https://github.com/tox-dev/sphinx-autodoc-typehints/issues/454

Maybe this PR will trigger some pip deps lock file magic, magic I'll have to do it in the CI. Gotta start somewhere.


chrismeyersfsu:
I'm like a monkey with a hand grenade. Closing this PR. I think this change needs to happen in `awx_plugins.interfaces` ???

PR #63, opened 2024-12-02
Fix docs ci

chrismeyersfsu:


chrismeyersfsu:
My attempt at manually re-creating the pip-tools ci workflow locally. I got one distro. Instead, I figured out how to point the `pip-tools` github ci workflow job at my branch that had the added constraint and to take it into account and regenerate the pip lock files.

PR #71, opened 2024-12-13
Add client_id and client_secret fields to Insights credential

fosterseth:
Adds `client_id` and `client_secret` fields to the existing insights credential so that users can authenticate with a service account instead of basic auth

so users should either use username/password, or client_id/client_secret

if client_id is set, then authentication method is set to "service_account" automatically

### DIFF
--- a/src/awx_plugins/credentials/plugins.py
+++ b/src/awx_plugins/credentials/plugins.py
@@ -586,16 +586,16 @@ insights = ManagedCredentialType(
     inputs={
         'fields': [
             {
-                'id': 'username', 'label': gettext_noop('Username'), 'type': 'string',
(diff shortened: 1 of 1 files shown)


djyasin (approved):

chrismeyersfsu:
Looks like this will work if your plan is to consume it in the project update, but I _think_ you will need to define a `custom_injectors` for this to work for the inventory update case.

edit: Actually, maybe not. But something over in [inventory plugin land](https://github.com/ansible/awx-plugins/blob/devel/src/awx_plugins/inventory/plugins.py#L428) will need to be updated to produce 

```
plugin: redhat.insights.insights

authentication: basic
user: "insights username"
password: "insights password"
```

OR

```
plugin: redhat.insights.insights
authentication: service_account
client_id: "service account client-id"
client_secret: "service account client-secret"
```

chrismeyersfsu (approved):

PR #87, opened 2025-02-19
Add a credential plugin that uses GitHub Apps to get tokens

webknjaz:
This patch uses PyGitHub to implement it. It currently only supports
GitHub App IDs due to limitations in PyGitHub [[1]]. It also
takes care of passing it as a string into the library because PyGitHub
does not perform type conversion, which sometimes leads to producing
invalid JWTs [[2]] as the underlying library PyJWT lacks validation
[[3]].

This change temporarily excludes mentions of client ID and marks it as
not allowed.

The plugin accepts both GitHub App and installation IDs as integers
and normalizes them into strings.

The tests use ephemerally-produced in-memory RSA keys for JWT
verification. As an optimization, they are 1024-bit sized because it
is cheaper and makes the tests more performant compared to bigger
values.

As a temporary measure, the patch includes a suppression of WPS202
in flake8 for `github_app.py`.

A few type-ignores were added around PyGitHub exception handling due
to them using `Any` [[4]].

The patch also includes a change to the flake8-eradicate config
letting distinguish comments with URL lists.

Supersedes #84.

[1]: https://github.com/PyGithub/PyGithub/issues/3213
[2]: https://github.com/PyGithub/PyGithub/issues/3214
[3]: https://github.com/jpadilla/pyjwt/issues/1039
[4]: https://github.com/PyGithub/PyGithub/issues/3218

Co-authored-by: Sviatoslav Sydorenko <webknjaz@redhat.com>


djyasin (approved):

fosterseth (approved):

src/awx_plugins/credentials/github_app.py
```diff
@@ -0,0 +1,267 @@
...
+
+github_app_inputs: GitHubAppInputs = {
+    'fields': [
+        {
+            'id': 'github_api_url',
+            'label': _('GitHub API endpoint URL'),
+            'type': 'string',
+            'help_text': _(
+                'Specify the GitHub API URL here. In the case of an Enterprise: '
+                ' https://gh.your.org/api/v3 (self-hosted) '
```
webknjaz: (double whitespace)
```suggestion
                'https://gh.your.org/api/v3 (self-hosted) '
```

src/awx_plugins/credentials/github_app.py
```diff
@@ -0,0 +1,267 @@
...
+        {
+            'id': 'install_id',
+            'label': _('GitHub App Installation ID'),
+            'type': 'string',
+            'help_text': _(
+                'The Installation ID from the GitHub App installation '
+                'generated by the GitHub Admin. '
+                'Example: 59980338 extracted from the installation link: '
+                'https://github.com/settings/installations/59980338 '
+                'required for creating a limited GitHub app token',
```
webknjaz:
```suggestion
                'required for creating a limited GitHub app token.',
```

src/awx_plugins/credentials/github_app.py
```diff
@@ -0,0 +1,267 @@
...
+        },
+        {
+            'id': 'app_or_client_id',
+            'label': _('GitHub App ID'),
+            'type': 'string',
+            'help_text': _(
+                'The GitHub App ID created by the GitHub Admin. '
+                'Example App ID: 1121547 '
+                'found on https://github.com/settings/apps/ '
+                'required for creating a JWT token for authentication',
```
webknjaz:
```suggestion
                'required for creating a JWT token for authentication.',
```

src/awx_plugins/credentials/github_app.py
```diff
@@ -0,0 +1,267 @@
...
+            ),
+        },
+        {
+            'id': 'install_id',
+            'label': _('GitHub App Installation ID'),
+            'type': 'string',
+            'help_text': _(
+                'The Installation ID from the GitHub App installation '
+                'generated by the GitHub Admin. '
+                'Example: 59980338 extracted from the installation link: '
```
webknjaz: ?
```suggestion
                'Example: 59980338 extracted from the installation link '
```

arrestle (approved):

fosterseth (approved):

PR #106, opened 2025-04-14
Add plugin for Azure VM managed identity

jessicamack:
This PR adds a plugin that uses Managed Identities for Azure resources on an Azure VM to authenticate and retrieve secrets from an Azure Key Vault. This is similar to how the existing Azure Key Vault plugin works but it does not require Client ID, Client Secret, or Tenant ID. When run in an Azure VM the identity of the VM will be picked up and used as the credential to authenticate to the Key Vault. The only requirements then are the Vault URL, the name of the secret to retrieve, optionally the version of the secret to retrieve, and for the VM itself to be given the correct Key Vault Administrator role.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -7,6 +7,11 @@ from importlib.metadata import version as _retrieve_metadata_version_for
 from pathlib import Path
 from tomllib import loads as _parse_toml
 
+from docutils.nodes import literal, reference
+from sphinx.addnodes import pending_xref
+from sphinx.application import Sphinx
+from sphinx.environment import BuildEnvironment
+
 
 # -- Path setup --------------------------------------------------------------
 
@@ -226,3 +231,42 @@ nitpick_ignore = [
     ),
     ('py:class', 'EnvVarsType'),
 ]
+
+
+def _replace_missing_tokencredential_reference(
+    app: Sphinx,
+    env: BuildEnvironment,
+    node: pending_xref,
+    contnode: literal,
+) -> reference | None:
+    if (node.get('refdomain'), node.get('reftype')) != ('py', 'class'):
+        return None
+
+    ref_target = node.get('reftarget', '')
+    if ref_target != 'azure.core.credentials.TokenCredential':
+        return None
+
+    return reference(
+        ref_target,
+        ref_target,
+        internal=False,
+        refuri='https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/core/azure-core/azure/core/credentials.py',
+    )
+
+
+def setup(app: Sphinx) -> dict[str, bool | str]:
+    """Register project-local Sphinx extension-API customizations.
+
+    :param app: Initialized Sphinx app instance.
+    :returns: Extension metadata.
+    """
+    app.connect(
+        'missing-reference',
+        _replace_missing_tokencredential_reference,
+    )
+
+    return {
+        'parallel_read_safe': True,
+        'parallel_write_safe': True,
+        'version': release,
+    }
--- a/src/awx_plugins/credentials/azure_kv.py
+++ b/src/awx_plugins/credentials/azure_kv.py
@@ -1,5 +1,5 @@
 # FIXME: the following violations must be addressed gradually and unignored
-# mypy: disable-error-code="import-untyped, no-untyped-def"
+# mypy: disable-error-code="no-untyped-def"
 
 from awx_plugins.interfaces._temporary_private_django_api import (  # noqa: WPS436
     gettext_noop as _,
@@ -7,7 +7,10 @@ from awx_plugins.interfaces._temporary_private_django_api import (  # noqa: WPS4
 
 from azure.identity import ClientSecretCredential
 from azure.keyvault.secrets import SecretClient
-from msrestazure import azure_cloud
+# NOTE: `msrestazure` is deprecated and does not provide usable typing.
+# NOTE: This suppression should be replaced by in-tree type stubs.
+# NOTE: Alternatively, the dependency can be replaced.
+from msrestazure import azure_cloud  # type: ignore[import-untyped]
 
 from .plugin import CredentialPlugin
 
--- a/src/awx_plugins/credentials/azure_kv.py
+++ b/src/awx_plugins/credentials/azure_kv.py
@@ -1,12 +1,31 @@
-# FIXME: the following violations must be addressed gradually and unignored
-# mypy: disable-error-code="no-untyped-def"
+"""Microsoft Azure Key Vault Lookup Plugin.
+
+This module defines a credential lookup plugin to authenticate and retrieve
+secrets from an Azure Key Vault. If the Client ID, Tenant ID, and Client
+Secret are provided it will create a credential with those. If one is missing,
+it will attempt to use the Managed Identity of an Azure VM to create a
+credential.
+
+Functions:
+
+- :func:`azure_keyvault_backend`: Creates a credential either with the fields
+  provided or via the VM environment, and retrieves the secret from the Key
+  Vault.
+- ``azure_keyvault_plugin``: Defines the credential plugin interface.
+"""
 
 from awx_plugins.interfaces._temporary_private_django_api import (  # noqa: WPS436
     gettext_noop as _,
 )
 
-from azure.identity import ClientSecretCredential
+from azure.core.credentials import TokenCredential
+from azure.identity import (
+    ClientSecretCredential,
+    CredentialUnavailableError,
+    ManagedIdentityCredential,
+)
 from azure.keyvault.secrets import SecretClient
+
 # NOTE: `msrestazure` is deprecated and does not provide usable typing.
 # NOTE: This suppression should be replaced by in-tree type stubs.
 # NOTE: Alternatively, the dependency can be replaced.
@@ -75,25 +94,69 @@ azure_keyvault_inputs = {
     ],
     'required': [
         'url',
-        'client',
-        'secret',
-        'tenant',
         'secret_field',
     ],
 }
 
 
-def azure_keyvault_backend(**kwargs):
-    csc = ClientSecretCredential(
-        tenant_id=kwargs['tenant'],
-        client_id=kwargs['client'],
-        client_secret=kwargs['secret'],
-    )
-    kv = SecretClient(credential=csc, vault_url=kwargs['url'])
-    return kv.get_secret(
-        name=kwargs['secret_field'],
-        version=kwargs.get('secret_version', ''),
-    ).value
+def _initialize_credential(
+    tenant: str = '',
+    client: str = '',
+    secret: str = '',
+) -> TokenCredential:
+    explicit_credentials_provided = all((tenant, client, secret))
+
+    if explicit_credentials_provided:
+        return ClientSecretCredential(
+            tenant_id=tenant,
+            client_id=client,
+            client_secret=secret,
+        )
+
+    return ManagedIdentityCredential()
+
+
+# WPS211 "too many args" is controlled externally
+# pylint: disable-next=too-many-arguments,too-many-positional-arguments
+def azure_keyvault_backend(  # noqa: WPS211
+    *,
+    url: str,
+    client: str,
+    secret: str,
+    tenant: str,
+    secret_field: str,
+    secret_version: str,
+) -> str | None:
+    """Get a credential and retrieve a secret from an Azure Key Vault.
+
+    An empty string for an optional parameter counts as not provided.
+
+    :param url: An Azure Key Vault URI.
+    :param client: The Client ID  (optional).
+    :param secret: The Client Secret  (optional).
+    :param tenant: The Tenant ID  (optional).
+    :param secret_field: The name of the secret to retrieve from the
+        vault.
+    :param secret_version: The version of the secret to retrieve
+        (optional).
+    :returns: The secret from the Key Vault.
+    :raises RuntimeError: If the software is not being run on an Azure
+        VM.
+    """
+    chosen_credential = _initialize_credential(tenant, client, secret)
+    keyvault = SecretClient(credential=chosen_credential, vault_url=url)
+    try:
+        keyvault_secret = keyvault.get_secret(
+            name=secret_field,
+            version=secret_version,
+        )
+    except CredentialUnavailableError as secret_lookup_err:
+        raise RuntimeError(
+            'You are not operating on an Azure VM, so the Managed Identity '
+            'feature is unavailable. Please provide the full Client ID, '
+            'Client Secret, and Tenant ID or run the software on an Azure VM.',
+        ) from secret_lookup_err
+    return keyvault_secret.value
 
 
 azure_keyvault_plugin = CredentialPlugin(
--- a/tests/azure_kv_test.py
+++ b/tests/azure_kv_test.py
@@ -1,0 +1,80 @@
+"""Tests Azure Key Vault credential plugin."""
+
+import pytest
+
+from azure.keyvault.secrets import (
+    KeyVaultSecret,
+    SecretClient,
+    SecretProperties,
+)
+
+from awx_plugins.credentials import azure_kv
+
+
+class _FakeSecretClient(SecretClient):
+    def get_secret(
+        self: '_FakeSecretClient',
+        name: str,
+        version: str | None = None,
+        **kwargs: str,
+    ) -> KeyVaultSecret:
+        props = SecretProperties()
+        return KeyVaultSecret(properties=props, value='test-secret')
+
+
+def test_azure_kv_invalid_env() -> None:
+    """Test running outside of Azure raises error."""
+    error_msg = (
+        'You are not operating on an Azure VM, so the Managed Identity '
+        'feature is unavailable. Please provide the full Client ID, '
+        'Client Secret, and Tenant ID or run the software on an Azure VM.'
+    )
+
+    with pytest.raises(
+        RuntimeError,
+        match=error_msg,
+    ):
+        azure_kv.azure_keyvault_backend(
+            url='https://test.vault.azure.net',
+            client='',
+            secret='client-secret',
+            tenant='tenant-id',
+            secret_field='secret',
+            secret_version='',
+        )
+
+
+@pytest.mark.parametrize(
+    ('client', 'secret', 'tenant'),
+    (
+        pytest.param('', '', '', id='managed-identity'),
+        pytest.param(
+            'client-id',
+            'client-secret',
+            'tenant-id',
+            id='client-secret-credential',
+        ),
+    ),
+)
+def test_azure_kv_valid_auth(
+    monkeypatch: pytest.MonkeyPatch,
+    client: str,
+    secret: str,
+    tenant: str,
+) -> None:
+    """Test successful Azure authentication via Managed Identity and credentials."""
+    monkeypatch.setattr(
+        azure_kv,
+        'SecretClient',
+        _FakeSecretClient,
+    )
+
+    keyvault_secret = azure_kv.azure_keyvault_backend(
+        url='https://test.vault.azure.net',
+        client=client,
+        secret=secret,
+        tenant=tenant,
+        secret_field='secret',
+        secret_version='',
+    )
+    assert keyvault_secret == 'test-secret'


src/awx_plugins/credentials/azure_mi.py
```diff
@@ -0,0 +1,60 @@
+# FIXME: the following violations must be addressed gradually and unignored
+# mypy: disable-error-code="import-untyped, no-untyped-def"
```
webknjaz: Try to avoid having ignored checks in freshly added modules. This was a band-aid for pre-existing code, but for new code, it shouldn't be allowed to have such global ignores. If there's places that generate errors, add `# type: ignore[error-code]` inline and a comment explaining what's wrong and when it can be dropped. Global ignores allow more errors to sneak in over time.

.mypy.ini
```diff
@@ -69,6 +69,10 @@ disallow_any_expr = false
 # crashes with some decorators like `@functools.cache`:
 disallow_any_expr = false
 
+[mypy-awx_plugins.credentials.azure_mi]
+# crashes with some decorators like `@functools.cache`:
```
webknjaz: Unless, there's an actual error violation that's happening, let's not disable rules globally. And if there are, it's best to inline ignores on specific lines, with explanatory comments.

src/awx_plugins/credentials/azure_mi.py
```diff
@@ -0,0 +1,60 @@
...
+        },
+    ],
+    'required': [
+        'url',
+        'secret_field',
+    ],
+}
+
+
+def azure_managed_id_backend(**kwargs):
```
webknjaz: Looks like this can be unpacked into just 3 args? Let's do this for the sake of proper typing and simplicity.

src/awx_plugins/credentials/azure_mi.py
```diff
@@ -0,0 +1,60 @@
...
+    credential = ManagedIdentityCredential()
+    kv = SecretClient(credential=credential, vault_url=kwargs['url'])
```
webknjaz: Could you change `kv` to something that would communicate what the object is?

src/awx_plugins/credentials/azure_mi.py
jessicamack: I've merged the new work into an existing plugin and tried to address the linting issues in that plugin.

pyproject.toml
```diff
@@ -152,6 +152,11 @@ credentials-azure-kv = [
   "azure-keyvault",
   "msrestazure",
 ]
+credentials-azure-mi = [
```
webknjaz: is a separate extra still needed?

src/awx_plugins/credentials/azure_kv.py
```diff
@@ -1,11 +1,26 @@
...
+This module defines a credential lookup plugin to authenticate and retrieve
+secrets from an Azure Key Vault. If the Client ID, Tenant ID, and Client Secret
+are provided it will create a credential with those. If one is missing, it will attemmpt
+to use the Managed Identity of an Azure VM to create a credential.
+
+Functions:
+
+- :func:`azure_keyvault_backend`: Creates a credential either with the fields
+provided or via the VM environment, and retrieves the secret from the Key Vault.
+- ``azure_keyvault_plugin``: Defines the credential plugin interface.
```
webknjaz, line 14:
```console
 /home/docs/checkouts/readthedocs.org/user_builds/awx-plugins-core/checkouts/106/src/awx_plugins/credentials/azure_kv.py:docstring of awx_plugins.credentials.azure_kv:11: WARNING: Bullet list ends without a blank line; unexpected unindent. [docutils] 
```
*(https://app.readthedocs.org/projects/awx-plugins-core/builds/27888342/#267943005--42)*

I think, it wants this:
```suggestion
- ``azure_keyvault_plugin``: Defines the credential plugin interface.

```

src/awx_plugins/credentials/azure_kv.py
```diff
@@ -71,20 +86,41 @@
...
+    """Get a credential and retrieve a secret from an Azure Key Vault.
+
+    :param url: An Azure Key Vault URI.
+    :param client (optional): The Client ID.
+    :param secret (optional): The Client Secret.
+    :param tenant (optional): The Tenant ID.
+    :param secret_field: The name of the secret to retrieve from the
+        vault.
+    :param secret_version (optional): The version of the secret to
+        retrieve.
```
webknjaz: invalid syntax
```suggestion
    :param client: The Client ID (optional).
    :param secret: The Client Secret (optional).
    :param tenant: The Tenant ID (optional).
    :param secret_field: The name of the secret to retrieve from the
        vault.
    :param secret_version: The version of the secret to
        retrieve (optional).
```

src/awx_plugins/credentials/azure_kv.py
```diff
@@ -71,20 +86,41 @@
...
 }
 
 
-def azure_keyvault_backend(**kwargs):
-    csc = ClientSecretCredential(
-        tenant_id=kwargs['tenant'],
-        client_id=kwargs['client'],
-        client_secret=kwargs['secret'],
-    )
+def azure_keyvault_backend(**kwargs: str) -> str:
```
webknjaz: Invalid typing. `kwargs` here is a dict, it'll be difficult to type it. It's best to list the individual args instead.

src/awx_plugins/credentials/azure_kv.py
```diff
@@ -71,20 +86,41 @@
...
+    :param secret_field: The name of the secret to retrieve from the
+        vault.
+    :param secret_version (optional): The version of the secret to
+        retrieve.
+    :returns: The secret from the Key Vault.
+    :raises: RuntimeError: If the software is not being run on an Azure
+        VM.
+    """
+    try:
+        if '' in [kwargs['tenant'], kwargs['client'], kwargs['secret']]:
```
webknjaz: Which of these lines is expected to raise an exception? It's an antipattern to have more than one instruction in a try-block.

src/awx_plugins/credentials/azure_kv.py
```diff
@@ -71,20 +86,41 @@
...
+                tenant_id=kwargs['tenant'],
+                client_id=kwargs['client'],
+                client_secret=kwargs['secret'],
+            )
+    except CredentialUnavailableError:
+        raise RuntimeError(
+            'You are not operating on an Azure VM, so the Managed Identity '
+            'feature is unavailable. Please provide the full Client ID, '
+            'Client Secret, and Tenant ID or run the software on an Azure VM.',
+        )
```
webknjaz: add a `.. from cred_err`

src/awx_plugins/credentials/azure_mi.py
jessicamack: yes, changing `csc` to `cred` and `kv` to `keyvault`

pyproject.toml
jessicamack: no, I missed this and one other change from the azure_mi work. removing both now

src/awx_plugins/credentials/azure_kv.py
```diff
@@ -71,20 +86,41 @@
...
+    :param url: An Azure Key Vault URI.
+    :param client (optional): The Client ID.
+    :param secret (optional): The Client Secret.
+    :param tenant (optional): The Tenant ID.
+    :param secret_field: The name of the secret to retrieve from the
+        vault.
+    :param secret_version (optional): The version of the secret to
+        retrieve.
+    :returns: The secret from the Key Vault.
+    :raises: RuntimeError: If the software is not being run on an Azure
```
webknjaz:
```suggestion
    :raises RuntimeError: If the software is not being run on an Azure
```

src/awx_plugins/credentials/azure_kv.py
jessicamack: https://github.com/ansible/awx-plugins/blob/fdfe0f6df76715ead11b9b7fc25181b56cf0f163/src/awx_plugins/credentials/azure_kv.py#L111 will raise the exception if the code is not run on an Azure VM but we should only try to use ManagedIdentityCredential if one of the variables in that list is missing

src/awx_plugins/credentials/azure_kv.py
jessicamack: I've listed the individual args.

src/awx_plugins/credentials/azure_kv.py
```diff
@@ -1,11 +1,28 @@
-# FIXME: the following violations must be addressed gradually and unignored
-# mypy: disable-error-code="import-untyped, no-untyped-def"
+"""Microsoft Azure Key Vault Lookup Plugin.
+
+This module defines a credential lookup plugin to authenticate and retrieve
+secrets from an Azure Key Vault. If the Client ID, Tenant ID, and Client
+Secret are provided it will create a credential with those. If one is missing,
+ it will attemmpt to use the Managed Identity of an Azure VM to create a
```
webknjaz: I think RST might be unhappy about this indent/leading whitespace, as it starts the block quote.
```suggestion
it will attempt to use the Managed Identity of an Azure VM to create a
```

src/awx_plugins/credentials/azure_kv.py
```diff
@@ -71,24 +88,52 @@
...
-    csc = ClientSecretCredential(
-        tenant_id=kwargs['tenant'],
-        client_id=kwargs['client'],
-        client_secret=kwargs['secret'],
-    )
-    kv = SecretClient(credential=csc, vault_url=kwargs['url'])
-    return kv.get_secret(
-        name=kwargs['secret_field'],
-        version=kwargs.get('secret_version', ''),
+def azure_keyvault_backend(
```
webknjaz:
```suggestion
# WPS211 "too many args" is controlled externally
# pylint: disable-next=too-many-arguments,too-many-positional-arguments
def azure_keyvault_backend(  # noqa: WPS211
```

src/awx_plugins/credentials/azure_kv.py
```diff
@@ -71,24 +88,52 @@
...
+        client: str,
+        secret: str,
+        tenant: str,
+        secret_field: str,
+        secret_version: str,
+) -> str:
+    """Get a credential and retrieve a secret from an Azure Key Vault.
+
+    :param url: An Azure Key Vault URI.
+    :param client: The Client ID  (optional).
```
webknjaz, line 135: What does “optional” mean here? Can it be `None`?

src/awx_plugins/credentials/azure_kv.py
```diff
@@ -71,24 +88,52 @@
...
-        tenant_id=kwargs['tenant'],
-        client_id=kwargs['client'],
-        client_secret=kwargs['secret'],
-    )
-    kv = SecretClient(credential=csc, vault_url=kwargs['url'])
-    return kv.get_secret(
-        name=kwargs['secret_field'],
-        version=kwargs.get('secret_version', ''),
+def azure_keyvault_backend(
+        url: str,
```
webknjaz: Let's make all the args keyword-only since that's the only way awx invokes this. Plus, it's usually bad to allow positional args when there's many.
```suggestion
        *,
        url: str,
```

src/awx_plugins/credentials/azure_kv.py
```diff
@@ -71,24 +88,52 @@
...
+    :param secret_field: The name of the secret to retrieve from the
+        vault.
+    :param secret_version: The version of the secret to retrieve
+        (optional).
+    :returns: The secret from the Key Vault.
+    :raises RuntimeError: If the software is not being run on an Azure
+        VM.
+    """
+    try:
+        if '' in [tenant, client, secret]:
```
webknjaz:
```suggestion
        explicit_credentials_empty = '' in {tenant, client, secret}:
        if explicit_credentials_empty:
```

src/awx_plugins/credentials/azure_kv.py
```diff
@@ -71,24 +88,52 @@
...
+            cred = ManagedIdentityCredential()
```
webknjaz: Since only this line is expected to raise that exception, move the try-block to only wrap this line but not others.

src/awx_plugins/credentials/azure_kv.py
```diff
@@ -1,11 +1,28 @@
...
 )
 
-from azure.identity import ClientSecretCredential
+from azure.identity import (
+    ClientSecretCredential,
+    CredentialUnavailableError,
+    ManagedIdentityCredential,
+)
 from azure.keyvault.secrets import SecretClient
 from msrestazure import azure_cloud
```
webknjaz: Technically, removing the global MyPy ignores at the top is out of the scope of this PR. But if you're doing that, here's how you can suppress missing typing in this module:
```suggestion
# NOTE: `msrestazure` is deprecated and does not provide usable typing.
# NOTE: This suppression should be replaced by in-tree type stubs.
# NOTE: Alternatively, the dependency can be replaced.
from msrestazure import azure_cloud  # type: ignore[import-untyped]
```
The project is deprecated, so I don't think anybody will ever ship installable typing info for it.

Another solution (cleaner, if you ask me) would be to create own types stubs and host them in-repo. Like I did here: https://github.com/ansible/awx_plugins.interfaces/tree/devel/_type_stubs. This will avoid the need to suppress things. Though, this can be done in a follow-up.

src/awx_plugins/credentials/azure_kv.py
```diff
@@ -71,24 +88,52 @@
...
+        (optional).
+    :returns: The secret from the Key Vault.
+    :raises RuntimeError: If the software is not being run on an Azure
+        VM.
+    """
+    try:
+        if '' in [tenant, client, secret]:
+            cred = ManagedIdentityCredential()
+        else:
+            cred = ClientSecretCredential(
```
webknjaz: The first `cred` assignment makes MyPy infer the type of the variable as `ManagedIdentityCredential`. And so it complains about seeing a differently typed thing being assigned here. This is because `ClientSecretCredential` is not a subclass of `ManagedIdentityCredential`.

You need to explicitly state the type where the variable is first defined (or in a separate instruction before the first use).
You could state that it's a union and add `: ClientSecretCredential | ManagedIdentityCredential` there. However, I checked the source code @ https://github.com/Azure/azure-sdk-for-python/blob/17b3e200/sdk/keyvault/azure-keyvault-administration/azure/keyvault/administration/_internal/client_base.py#L70C52-L70C67 and discovered that `SecretClient(cred=)` accepts anything shaped as a `TokenCredential` which is defined as a protocol.

So a better solution would be to use `cred: TokenCredential` instead (I'd use a longer var name than `cred`, though). Doing `from azure.core.credentials import TokenCredential` at the top should make it available for use.

src/awx_plugins/credentials/azure_kv.py
```diff
@@ -71,24 +88,52 @@
...
-        name=kwargs['secret_field'],
-        version=kwargs.get('secret_version', ''),
+def azure_keyvault_backend(
+        url: str,
+        client: str,
+        secret: str,
+        tenant: str,
+        secret_field: str,
+        secret_version: str,
+) -> str:
```
webknjaz: MyPy says that `.value` can also be `None`. So the correct return type is
```suggestion
) -> str | None:
```

If you want/need to keep it `str`, you'll need to make sure you don't ever return `None`.

src/awx_plugins/credentials/azure_kv.py
jessicamack: if a Client ID is not provided I believe an empty string will be submitted and a Client ID is not required

webknjaz:
Unrelated linkcheck failure: #107

src/awx_plugins/credentials/azure_kv.py
```diff
@@ -1,13 +1,35 @@
...
     gettext_noop as _,
 )
 
-from azure.identity import ClientSecretCredential
+from azure.identity import (
+    ClientSecretCredential,
+    CredentialUnavailableError,
+    ManagedIdentityCredential,
+    TokenCredential,
+)
```
webknjaz:
```suggestion
from azure.identity import (
    ClientSecretCredential,
    CredentialUnavailableError,
    ManagedIdentityCredential,
)
from azure.core.credentials import TokenCredential
```
*(https://github.com/ansible/awx-plugins/pull/106#discussion_r2049171130)*
*(https://github.com/ansible/awx-plugins/actions/runs/14603464110/job/40967062189?pr=106#step:13:75)*

src/awx_plugins/credentials/azure_kv.py
webknjaz: Then I'd document that “empty string counts as not provided” or something. “Optional” is often interpreted as `None` since it's a popular sentinel value, but in this case it sounds like an empty string is a sentinel.

src/awx_plugins/credentials/azure_kv.py
```diff
@@ -71,24 +93,64 @@
...
-    csc = ClientSecretCredential(
-        tenant_id=kwargs['tenant'],
-        client_id=kwargs['client'],
-        client_secret=kwargs['secret'],
-    )
-    kv = SecretClient(credential=csc, vault_url=kwargs['url'])
-    return kv.get_secret(
-        name=kwargs['secret_field'],
-        version=kwargs.get('secret_version', ''),
+def _initialize_credential(tenant: str = '', client: str = '', secret='', /) -> TokenCredential:
```
webknjaz:
```suggestion
def _initialize_credential(tenant: str = '', client: str = '', secret: str = '', /) -> TokenCredential:
```

src/awx_plugins/credentials/azure_kv.py
```diff
@@ -71,24 +93,71 @@
...
-    )
-    kv = SecretClient(credential=csc, vault_url=kwargs['url'])
-    return kv.get_secret(
-        name=kwargs['secret_field'],
-        version=kwargs.get('secret_version', ''),
+def _initialize_credential(
+    tenant: str = '',
+    client: str = '',
+    secret: str = '',
+    /,
```
webknjaz: I made a mistake in the initial suggested signature: https://wemake-python-styleguide.readthedocs.io/en/latest/pages/usage/violations/best_practices.html#wemake_python_styleguide.violations.best_practices.ProblematicFunctionParamsViolation. It doesn't really make sense to have a bunch of positional-only params with defaults.
```suggestion
```

docs/conf.py
```diff
@@ -226,3 +231,44 @@
...
+
+    try:
+        return reference(
+            ref_target,
+            ref_target,
+            internal=False,
+            refuri='https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/core/azure-core/azure/core/credentials.py',
+        )
+    except KeyError:
+        return None
```
webknjaz: This is never going to happen. Drop it.

docs/conf.py
```diff
@@ -226,3 +231,44 @@
...
+
+
+def setup(app: Sphinx) -> dict[str, bool | str]:
+    """Register project-local Sphinx extension-API customizations.
+    :param app: Initialized Sphinx app instance.
+    :type app: Sphinx
```
webknjaz: Type info isn't needed here since the annotations cover it already.

docs/conf.py
```diff
@@ -226,3 +231,44 @@
...
+    :returns: Extension metadata.
+    :rtype: dict[str, bool | str]
```
webknjaz: This too.

webknjaz (commented):
Found a few things to delete.

docs/conf.py
```diff
@@ -244,24 +244,15 @@ def _replace_missing_tokencredential_reference(
...
+        ref_target,
+        ref_target,
+        internal=False,
+        refuri='https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/core/azure-core/azure/core/credentials.py',
+    )
 
 
 def setup(app: Sphinx) -> dict[str, bool | str]:
-    """Register project-local Sphinx extension-API customizations.
-    :param app: Initialized Sphinx app instance.
```
webknjaz: @jessicamack param descriptions might still be useful. It's the types that don't need to be duplicated.

docs/conf.py
```diff
@@ -226,3 +231,36 @@
...
+        return None
+
+    ref_target = 'azure.core.credentials.TokenCredential'
+
+    return reference(
+        ref_target,
+        ref_target,
+        internal=False,
+        refuri='https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/core/azure-core/azure/core/credentials.py',
+    )
```
webknjaz: Don't forget to also compare the ref target on the `node` object against the `ref_target` variable. Otherwise, it'll use this link for any `py:class` object with a missing reference if more of those appear in the future.

src/awx_plugins/credentials/azure_kv.py
```diff
@@ -72,24 +94,71 @@
...
+def azure_keyvault_backend(  # noqa: WPS211
+    *,
+    url: str,
+    client: str,
+    secret: str,
+    tenant: str,
+    secret_field: str,
+    secret_version: str,
+) -> str | None:
+    """Get a credential and retrieve a secret from an Azure Key Vault. An
```
webknjaz: The summary line is just a single short sentence. The description goes into a separate paragraph per PEP 257.

webknjaz (commented):
In addition to what the linter wants you to fix, could you also see if it's possible to produce some code coverage for the new lines in the diff?

I think that it's fairly easy to write a test for `_initialize_credential()` (it seems to just have 3 code branches). Use `monkeypatch.setenv()` and `monkeypatch.delenv()` to cover cases without explicitly passed args.

And for `azure_keyvault_backend()` it might be a bit trickier but not impossible. You may end up having to use the `monkeypatch.setattr()` or `mocker` / `mocker.spy()` in there.

src/awx_plugins/credentials/azure_kv.py
webknjaz: Also, `param` is not a complete word so the spell checker rightfully flags it.

djyasin (approved):

fosterseth (approved):

tvo318 (approved):
LGTM! 🚀

djyasin (approved):

fosterseth (approved):

tests/azure_kv_test.py
```diff
@@ -0,0 +1,95 @@
...
+            client='',
+            secret='client-secret',
+            tenant='tenant-id',
+            secret_field='secret',
+            secret_version='',
+        )
+
+
+@pytest.mark.parametrize(
+    ('azure_args'),
```
webknjaz: When it's just one arg, you don't need to wrap the string with parens.
```suggestion
    'azure_args',
```

tests/azure_kv_test.py
```diff
@@ -0,0 +1,95 @@
...
+)
+
+from awx_plugins.credentials import azure_kv
+
+
+class _FakeSecretClient(SecretClient):
+    def get_secret(
+        self: '_FakeSecretClient',
+        name: str,
+        version: Optional[str] = None,
```
webknjaz: Using `Optional` is considered confusing.
```suggestion
        version: str | None = None,
```

tests/azure_kv_test.py
```diff
@@ -0,0 +1,95 @@
+"""Tests Azure Key Vault credential plugin."""
+
+# ruff: noqa: S106 UP045
```
webknjaz: What are these? Every rule suppression must have a justification comment. Seeing alphanumeric identifiers is not enough to understand what the consequences are and what are the conditions for getting rid of temporary ignores.

tests/azure_kv_test.py
webknjaz: Although, it'd be more straightforward to just enumerate individual args that are parametrized and simplify typing, not even needing to have that `TypedDict`:
```suggestion
    ('client', 'secret', 'tenant'),
```

tests/azure_kv_test.py
```diff
@@ -0,0 +1,95 @@
...
+            secret_version='',
+        )
+
+
+@pytest.mark.parametrize(
+    ('azure_args'),
+    (
+        pytest.param(
+            {
+                'url': 'https://test.vault.azure.net',
```
webknjaz: Only put things that change in here. This param is the same, so it could go directly into the invocation. Duplicating things might be more difficult to follow.

tests/azure_kv_test.py
```diff
@@ -0,0 +1,95 @@
...
+                'client': '',
+                'secret': '',
+                'tenant': '',
+                'secret_field': 'secret',
+                'secret_version': '',
+            },
+            id='managed-identity-test',
```
webknjaz: I wouldn't add `test` to every param since the function name already has `test_` and it's also a part of the same test ID:
```suggestion
            id='managed-identity',
```

tests/azure_kv_test.py
```diff
@@ -0,0 +1,95 @@
...
+    secret: str
+    tenant: str
+    secret_field: str
+    secret_version: str
+
+
+def test_azure_kv_invalid_env() -> None:
+    """Test running outside of Azure raises error."""
+    with pytest.raises(
+        CredentialUnavailableError,
```
webknjaz: If this is happening, something is not being handled right in the implementation. I'd expect a `RuntimeError` and not underlying errors leaking into this layer. This is probably a bug somewhere that has to be fixed before merging.

tests/azure_kv_test.py
webknjaz: Looks like UP045 will be fixed with https://github.com/ansible/awx-plugins/pull/106/files#r2073709467 and S106 should be suppressed for all the tests and end up right after https://github.com/ansible/awx-plugins/blob/002bdc8/.ruff.toml#L155.

webknjaz: Addressed in #112 and #113.

webknjaz:
```suggestion
```

tests/azure_kv_test.py
```diff
@@ -0,0 +1,95 @@
+"""Tests Azure Key Vault credential plugin."""
+
+# ruff: noqa: S106 UP045
+from typing import Optional, TypedDict
```
webknjaz:
```suggestion
from typing import TypedDict
```

src/awx_plugins/credentials/azure_kv.py
```diff
@@ -72,24 +94,72 @@
...
+            client_id=client,
+            client_secret=secret,
+        )
+
+    try:
+        return ManagedIdentityCredential()
+    except CredentialUnavailableError as cred_err:
+        raise LookupError(
+            'Failed to detect credentials from environment.',
+        ) from cred_err
```
webknjaz: This is not covered by tests, so we don't know if it actually works. It might not be.
https://app.codecov.io/gh/ansible/awx-plugins/commit/94f7e3611fd08fcec49e22d81d6f7618ab8567fc/blob/src/awx_plugins/credentials/azure_kv.py?flags%5B0%5D=pytest&dropdown=coverage#L118

src/awx_plugins/credentials/azure_kv.py
```diff
@@ -72,24 +94,72 @@
...
+        VM.
+    """
+    try:
+        chosen_credential = _initialize_credential(tenant, client, secret)
+    except LookupError as cred_lookup_err:
+        raise RuntimeError(
+            'You are not operating on an Azure VM, so the Managed Identity '
+            'feature is unavailable. Please provide the full Client ID, '
+            'Client Secret, and Tenant ID or run the software on an Azure VM.',
+        ) from cred_lookup_err
```
webknjaz: https://app.codecov.io/gh/ansible/awx-plugins/commit/94f7e3611fd08fcec49e22d81d6f7618ab8567fc/blob/src/awx_plugins/credentials/azure_kv.py?flags%5B0%5D=pytest&dropdown=coverage#L153

tests/azure_kv_test.py
webknjaz: We have to figure out why it doesn't hit https://app.codecov.io/gh/ansible/awx-plugins/commit/94f7e3611fd08fcec49e22d81d6f7618ab8567fc/blob/src/awx_plugins/credentials/azure_kv.py?flags%5B0%5D=pytest&dropdown=coverage#L153

src/awx_plugins/credentials/azure_kv.py
```diff
@@ -72,24 +94,72 @@
...
+    if explicit_credentials_provided:
+        return ClientSecretCredential(
+            tenant_id=tenant,
+            client_id=client,
+            client_secret=secret,
+        )
+
+    try:
+        return ManagedIdentityCredential()
+    except CredentialUnavailableError as cred_err:
```
webknjaz: Looks like the assumption that initializing `ManagedIdentityCredential()` would raise `CredentialUnavailableError` was incorrect. I stepped through `test_azure_kv_invalid_env` and discovered that an instance of `ManagedIdentityCredential` is returned from this function w/o any errors. But then, `keyvault.get_secret()` in the caller is what raises the `CredentialUnavailableError` exception — that's where it's unhandled.

tests/azure_kv_test.py
webknjaz: *(also posted @ https://github.com/ansible/awx-plugins/pull/106/files#r2074049256)*

`CredentialUnavailableError` is not coming from the underlying `_initialize_credential()` call but from `keyvault.get_secret()`. This effectively means that the implementation needs some restructuring in order to fix the bug. And then, this test can be rewritten with correct expectations.

webknjaz (approved):

adrisala (commented):
LGTM, no regression observed 🚀

PR #109, opened 2025-04-22
🧪 Add Ruff linter to pre-commit

webknjaz:
It is integrated with auto-fixing enabled which effectively replaces `pyupgrade`, `isort` and `docformatter`. It does not yet fully [[1]] cover `autopep8`, though.

`autopep8` is being removed since the combination of `add-trailing-comma` and `ruff` works rather well with trailing commas.

And the formatter mode isn't enabled either due to partial incompatibility with `add-trailing-comma`.

The WPS flake8 plugin is upgraded to v1.1.0 as it drops a lot of checks performed by `ruff`. Other plugins are not removed and should be evaluated for being covered by `ruff` in follow-ups.

[1]: https://github.com/astral-sh/ruff/issues/5564#issuecomment-1623901448

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -37,38 +37,13 @@ repos:
   hooks:
   - id: add-trailing-comma
 
-- repo: https://github.com/asottile/pyupgrade.git
-  rev: v3.19.1
+- repo: https://github.com/astral-sh/ruff-pre-commit.git
+  rev: v0.11.6
   hooks:
-  - id: pyupgrade
+  - id: ruff
     args:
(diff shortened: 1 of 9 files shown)


webknjaz:
> ❌ Your patch check has failed because the patch coverage (31.25%) is below the target coverage (100.00%). You can increase the patch coverage or adjust the [target](https://docs.codecov.com/docs/commit-status#target) coverage.

This is fine (kinda) because the runtime changes only include linting fixes. It's not new code, but old pre-existing code that isn't adequately covered.

For this reason, I'm going to bypass this check.

PR #110, opened 2025-04-23
🧪 Integrate ruff formatting

webknjaz:
This patch adds a double-pass of `ruff format` to help it settle suggestions from `add-trailing-comma` during a single `pre-commit` run.
It also blame-ignores a dedicated formatting commit to help with future Git paleontology.


webknjaz:
Same as https://github.com/ansible/awx-plugins/pull/109#issuecomment-2821544626, force-merging.

PR #111, opened 2025-05-01
lower patch/pytest to 90 from 100

thedoubl3j:
* we as a team need to discuss what we want this to be, currently lowering to help unblock some work but a larger discussion is needed

This will become part of a larger effort to understand the CI of awx-plugins and adjust as needed to bring it in line with ATF.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -20,7 +20,7 @@ coverage:
       default:
         target: 100%
       pytest:
-        target: 100%
+        target: 90%
         flags:
         - pytest
       typing:


AlanCoding (approved):

djyasin (approved):

AlanCoding:
I thought the linter failure couldn't possibly be serious, but it appears it is:

```
nitpick check (check files only)..............................................Failed
- hook id: nitpick-check
- exit code: 1

.codecov.yml:1: NIP[36](https://github.com/ansible/awx-plugins/actions/runs/14777977382/job/41490437741?pr=111#step:14:37)9  has different values. Use this:
coverage:
  status:
    patch:
      pytest:
        target: 100%
Violations: ❌ 1 to fix manually.
```

Clearly, the suggestion would just undo this change.

Perhaps it means there is a duplicate value somewhere else (not stated in the error message) that is at 100% and needs to be 90% to have parity.

webknjaz:
@AlanCoding that's a check for configs having a certain required setting.

webknjaz:
In general, I don't think this is a good idea to accept this patch. It has already helped uncover bugs in implementation+tests of #106. This would be increasing the likelihood of sneaking more bugs in.

PR #116, opened 2025-05-08
Set defaults for optional values in azure_kv plugin

jessicamack:
The integration tests don't pass in an empty string like I assumed the UI does so tests are failing when optional arguments aren't given. This sets the default for the optional arguments to an empty string. I considered setting them to `str | None` but that introduced other typing issues.

### DIFF
--- a/src/awx_plugins/credentials/azure_kv.py
+++ b/src/awx_plugins/credentials/azure_kv.py
@@ -121,11 +121,11 @@ def _initialize_credential(
 def azure_keyvault_backend(  # noqa: WPS211
     *,
     url: str,
-    client: str,
-    secret: str,
-    tenant: str,
+    client: str = '',
+    secret: str = '',
+    tenant: str = '',
     secret_field: str,
-    secret_version: str,
+    secret_version: str = '',
 ) -> str | None:
     """Get a credential and retrieve a secret from an Azure Key Vault.
 


AlanCoding:
> [codecov/project](https://github.com/ansible/awx-plugins/pull/116/checks?check_run_id=41891581338) — 71.76% (-1.70%) compared to 85c96f9

Looks like a bad hit from `toxfile.py`, Codecov is giving me a 400 when I try to click into it.

AlanCoding (approved):
I expect this will fix the integration test errors.

src/awx_plugins/credentials/azure_kv.py
```diff
@@ -121,11 +121,11 @@ def _initialize_credential(
 def azure_keyvault_backend(  # noqa: WPS211
     *,
     url: str,
-    client: str,
-    secret: str,
-    tenant: str,
+    client: str = '',
+    secret: str = '',
+    tenant: str = '',
     secret_field: str,
```
webknjaz, line 127: Does it *actually* work with an argument that doesn't have a default in between the others? :thinking:

webknjaz: Oh.. That's probably because all these are kw-only.

webknjaz:
> The integration tests don't pass in an empty string like I assumed the UI does so tests are failing when optional arguments aren't given.

So you're suggesting that awx just doesn't include fields that might be empty otherwise? It'd be useful to link the CI run and maybe include a log excerpt to include the context here.

> This sets the default for the optional arguments to an empty string. I considered setting them to `str | None` but that introduced other typing issues.

What kind of typing issues? Any log? The only thing I can think of is that you might need to propagate those down to the helper function and maybe add a check for `is None` (or add an `assert ... is not None`) so that MyPy would perform type narrowing where it thinks the types would be incompatible. This is my best guess w/o seeing the log, though.

webknjaz:
In general, using `''` as a sentinel is probably fine. I'm wondering if #87 would have the same problem.

I'm also thinking that there might be a need to define the awx behavior more clearly. For example, it could look into the schema of plugin inputs and make sure to pass all the inputs listed. Having such a guarantee would make the whole thing more reliable.

webknjaz:
> > [codecov/project](https://github.com/ansible/awx-plugins/pull/116/checks?check_run_id=41891581338) — 71.76% (-1.70%) compared to [85c96f9](https://github.com/ansible/awx-plugins/commit/85c96f9d7b62321389fd72102fdc05c10196ce66)
> 
> Looks like a bad hit from `toxfile.py`, Codecov is giving me a 400 when I try to click into it.

Oh, I didn't realize this wasn't rebased. Although, Codecov was already taking it into account before.

I'll try to schedule a merge again and see what else might be happening..

webknjaz:
Also, I just double-verified that `codecov/project` isn't a required check. Only `codecov/patch/pytest` and `codecov/project/tests` are. And there's also a status from `pre-commit.ci`.

A few past `pre-commit.ci` runs for the merge queue attempts are green: https://results.pre-commit.ci/repo/github/836873755.

webknjaz:
I think this might be Codecov's flakiness — it sometimes gets stuck and reports its statuses rather late. And merge queues are configured to only wait for an hour and then fail with a timeout.

webknjaz:
So I was watching the merge queues widget and noticed the following required check failure before it dropped the attempt:
```console
 codecov/patch/pytest — 31.37% of diff hit (target 100.00%) 
```

This check corresponds to runtime coverage (not type checking) coming from pytest jobs. And its measurements are scoped to the `src/` directory.

I looked at https://app.codecov.io/gh/ansible/awx-plugins/commit/2703503d2009f2c32cdf26bc26e9dd94f1221a31/tree?search=&flags%5B0%5D=pytest, and I see that `src/` is shown as having `56.96%`.

Sometimes, Codecov starts computing and reporting coverage before it gets all the individual reports from different CI jobs. And these coverage checks may have low initial values but can then be updated as Codecov receives more files.
This shouldn't be happening, as I've set the minimum number of reports to be the number of the job in test matrices. But if the number of jobs changed and the expectation wasn't adjusted in the config, this may influence Codecov's behavior.

Also, merge queues cancel the attempt once they get the first required check failure. So I think it's possible that this is happening because Codecov is impatient...

webknjaz:
> `patch`

Oh, I missed that this is patch coverage.. Not the project one. It makes sense now. And `toxfile.py` does show up in there.

And `https://app.codecov.io/gh/ansible/awx-plugins/commit/2703503d2009f2c32cdf26bc26e9dd94f1221a31` does say `This commit 2703503 compared to 002bdc8`, which means it `diff`s against the two-week old 002bdc8df2af9f3a41b5e92808114a42320f0b9b. And this explains why `toxfile.py` isn't in the diff.

webknjaz:
It's weird that Codecov computes the diff based on such an old commit.

https://github.com/ansible/awx-plugins/commit/a809dbe5b6d9c578de40215a2264ba7544517328 clearly shows being rebased on top of `devel`.

And https://github.com/ansible/awx-plugins/commit/2703503d2009f2c32cdf26bc26e9dd94f1221a31 shows that its parents are `HEAD`s of this PR's branch and `devel` too.

Also, Codecov itself *knows* that there were newer commits on `devel` since and shows coverage for them just fine: https://app.codecov.io/gh/ansible/awx-plugins/commits/devel.

So I conclude that this is Codecov's backend inconsistency in that it mismatches things to diff.

I'll force-merge this PR, then.

jessicamack:
> > The integration tests don't pass in an empty string like I assumed the UI does so tests are failing when optional arguments aren't given.
> 
> So you're suggesting that awx just doesn't include fields that might be empty otherwise? It'd be useful to link the CI run and maybe include a log excerpt to include the context here.

the PR has been merged now but to answer your questions, the error Alan was seeing is `TypeError: azure_keyvault_backend() missing 1 required keyword-only argument: 'secret_version'`

> > This sets the default for the optional arguments to an empty string. I considered setting them to `str | None` but that introduced other typing issues.
> 
> What kind of typing issues? Any log? The only thing I can think of is that you might need to propagate those down to the helper function and maybe add a check for `is None` (or add an `assert ... is not None`) so that MyPy would perform type narrowing where it thinks the types would be incompatible. This is my best guess w/o seeing the log, though.

The typing issue was within the ClientSecretCredential method. I'm not sure how the linter wants this to be made more explicit. This is after also setting the kwargs in _initialize_credential to `str | None`
> src/awx_plugins/credentials/azure_kv.py:111:23:111:28: error: Argument "tenant_id" to "ClientSecretCredential" has incompatible type "str | None"; expected "str"  [arg-type]
                tenant_id=tenant,
                          ^~~~~~

webknjaz:
> > > The integration tests don't pass in an empty string like I assumed the UI does so tests are failing when optional arguments aren't given.
> > 
> > 
> > So you're suggesting that awx just doesn't include fields that might be empty otherwise? It'd be useful to link the CI run and maybe include a log excerpt to include the context here.
> 
> the PR has been merged now but to answer your questions, the error Alan was seeing is `TypeError: azure_keyvault_backend() missing 1 required keyword-only argument: 'secret_version'`

Sounds like a schema validation layer I suggested would be good to explore and implement.



> The typing issue was within the ClientSecretCredential method. I'm not sure how the linter wants this to be made more explicit. This is after also setting the kwargs in _initialize_credential to `str | None`
> 
> > src/awx_plugins/credentials/azure_kv.py:111:23:111:28: error: Argument "tenant_id" to "ClientSecretCredential" has incompatible type "str | None"; expected "str"  [arg-type]
> > tenant_id=tenant,
> > ^~~~~~

The `tenant: str | None` type persists across the function scope. MyPy knows that you're passing a potentially incompatible type into `ClientSecretCredential()` where all the arguments are typed as `str`: https://github.com/Azure/azure-sdk-for-python/blob/bd62a450f7667ccccc2eed91aee969807f39b468/sdk/identity/azure-identity/azure/identity/_credentials/client_secret.py#L43C24-L43C74. And `None` (a part of the `str | None` union) isn't compatible with `str`.

This is solved by making MyPy perform type narrowing: https://mypy.readthedocs.io/en/stable/type_narrowing.html.

There's a number of common patterns that MyPy recognizes to do this. There's if-blocks that check for types with `isinstance()`, for example. So if you were to do
```python
if tenant is None:
    return  # bail early

ClientSecretCredential(..., tenant_id=tenant, ...)
```
or
```python
if tenant is not None:
    ClientSecretCredential(..., tenant_id=tenant, ...)
```
or
```python
if not isinstance(tenant, str):
    return  # bail early

ClientSecretCredential(..., tenant_id=tenant, ...)
```
then, MyPy would change `tenant: str | None` to just `tenant: str` when analyzing what's passed into the initializer of `ClientSecretCredential` which would be compatible.

Another mechanism is using an `assert` statement. Here's an example from this repo's codebase: https://github.com/ansible/awx-plugins/blob/21173c4/src/awx_plugins/credentials/github_app.py#L163.

Because our implementation wasn't using a supported inspectable check, MyPy was unable to perform type narrowing automatically. Without rewriting the check style, we can force it as follows:
```diff
      explicit_credentials_provided = all((tenant, client, secret))

      if explicit_credentials_provided:
+         assert tenant is not None and client is not None and secret is not None
          return ClientSecretCredential(
              tenant_id=tenant,
              client_id=client,
              client_secret=secret,
          )
```

PR #119, opened 2025-06-11
Update credential injectors

pb82:
Based on #55 from @sean-m-sullivan

### DIFF
--- a/src/awx_plugins/credentials/plugins.py
+++ b/src/awx_plugins/credentials/plugins.py
@@ -754,7 +754,7 @@ controller = ManagedCredentialType(
                 'secret': False,
                 'default': 'galaxy',
                 'help_text': gettext_noop(
-                    'Specify the timeout Ansible should use in requests to'
+                    'Specify the timeout Ansible should use in requests to '
                     'the host. Defaults to 10s',
                 ),
             },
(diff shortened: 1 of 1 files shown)


webknjaz:
@pb82 FYI, your Git is likely misconfigured. GitHub doesn't link the commits to your account, which means they have a committer email that isn't known to GH. You should either verify it in your profile or update your Git configuration and amend the commits after.

src/awx_plugins/credentials/plugins.py
```diff
@@ -747,10 +747,29 @@
...
                 'secret': False,
             },
+            {
+                'id': 'request_timeout',
+                'label': gettext_noop('Verify SSL'),
+                'type': 'string',
+                'secret': False,
+                'default': 'galaxy',
+                'help_text': gettext_noop(
+                    'Specify the timeout Ansible should use in requests to'
```
webknjaz: Missing a whitespace in the implicit string concatenation. If you leave it like this, it'll render as `"tohost"` in the place where strings merge.
```suggestion
                    'Specify the timeout Ansible should use in requests to '
```

pb82: nice catch, thanks @webknjaz 
this is fixed now.

pb82:
@webknjaz sorry, new laptop. Didn't have my git user and email configured.

src/awx_plugins/credentials/plugins.py
webknjaz: @pb82 do you know you can just accept suggestions from GH UI?

webknjaz (approved):

PR #134, opened 2025-10-01
Add VMware ESXi inventory plugin

jessicamack:
Port the ESXi inventory plugin from tower to awx-plugins https://github.com/ansible/tower/pull/6895

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -121,6 +121,7 @@ satellite6 = "awx_plugins.inventory.plugins:satellite6_supported"
 controller = "awx_plugins.inventory.plugins:controller_supported"
 insights = "awx_plugins.inventory.plugins:insights_supported"
 openshift_virtualization = "awx_plugins.inventory.plugins:openshift_virtualization_supported"
+vmware_esxi = "awx_plugins.inventory.plugins:vmware_esxi_supported"
 
 [project.license]
 file = "LICENSE"
@@ -189,6 +190,10 @@ inventory-gce = [
   "PyYAML",
 ]
 inventory-vmware = [
+  "awx_plugins.interfaces",
+  "PyYAML",
+]
+inventory-vmware-esxi = [
   "awx_plugins.interfaces",
   "PyYAML",
 ]
--- a/src/awx_plugins/inventory/plugins.py
+++ b/src/awx_plugins/inventory/plugins.py
@@ -220,6 +220,7 @@ class vmware(PluginFileInjector):
 
 class vmware_esxi_supported(PluginFileInjector):
     plugin_name = 'esxi_hosts'
+    plugin_description = 'VMware ESXi'
     base_injector = 'managed'
     namespace = 'vmware'
     collection = 'vmware'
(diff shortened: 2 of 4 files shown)


pyproject.toml
```diff
@@ -121,6 +121,7 @@ satellite6 = "awx_plugins.inventory.plugins:satellite6_supported"
 controller = "awx_plugins.inventory.plugins:controller_supported"
 insights = "awx_plugins.inventory.plugins:insights_supported"
 openshift_virtualization = "awx_plugins.inventory.plugins:openshift_virtualization_supported"
+vmware_esxi = "awx_plugins.inventory.plugins:vmware_esxi"
```
webknjaz: Looks like all the other entry points in this section have a `_supported` suffix.

src/awx_plugins/inventory/plugins.py
```diff
@@ -218,6 +218,14 @@ class vmware(PluginFileInjector):
     collection = 'vmware'
 
 
+class vmware_esxi(PluginFileInjector):
+    plugin_name = 'esxi_hosts'
+    base_injector = 'managed'
+    namespace = 'vmware'
+    collection = 'vmware'
+    use_fqcn = True
```
webknjaz: Is this lacking a `plugin_description` attribute similar to `rhv_supported`?

tests/importable_test.py
```diff
@@ -94,6 +94,11 @@ def __str__(self) -> str:
         'vmware',
         'awx_plugins.inventory.plugins:vmware',
     ),
+    EntryPointParam(
+        'awx_plugins.inventory',
+        'vmware_esxi',
+        'awx_plugins.inventory.plugins:vmware_esxi',
+    ),
```
webknjaz: Looks like this this checks for things in `[project.entry-points."awx_plugins.inventory"]` and not in `[project.entry-points."awx_plugins.inventory.supported"]`. You might need a new list of params for that. Seems like nobody added that in the past.

pyproject.toml
jessicamack: realized this when I saw the failing test. the names have been updated

src/awx_plugins/inventory/plugins.py
jessicamack: added a description

tests/importable_test.py
jessicamack: further down it seems to check for the entries in `awx_plugins.inventory.supported`. I moved the new entry down with its like kind when I updated the name

pyproject.toml
```diff
@@ -121,6 +121,7 @@ satellite6 = "awx_plugins.inventory.plugins:satellite6_supported"
 controller = "awx_plugins.inventory.plugins:controller_supported"
 insights = "awx_plugins.inventory.plugins:insights_supported"
 openshift_virtualization = "awx_plugins.inventory.plugins:openshift_virtualization_supported"
+vmware_esxi = "awx_plugins.inventory.plugins:vmware_esxi_supported"
```
AlanCoding, line 124: I want to have my head on straight - _this_ is the string that will map to the namespace of the new CredentialType in AWX, is that correct? If so then :+1: that should match.

thedoubl3j: correct

thedoubl3j (approved):

PR #139, opened 2025-10-27
Add support for Iv2-prefixed Client IDs into Github App Credential

stevensonmichel:
As Github encourages more the use of Client IDs, one can create a Github App Installation Lookup App Credential using Github Client IDs starting with Iv1. However, using Iv2 will throw an error.

Fixes: as Github is providing Client IDs starting with Iv2, Credentials should be able to be created with it. The changes in this PR adds support for Iv2, enabling Credentials to be created using them.

### DIFF
--- a/src/awx_plugins/credentials/github_app.py
+++ b/src/awx_plugins/credentials/github_app.py
@@ -79,11 +79,11 @@ github_app_inputs: GitHubAppInputs = {
         },
         {
             'id': 'app_or_client_id',
-            'label': _('GitHub App ID'),
+            'label': _('GitHub App ID or Client ID'),
             'type': 'string',
             'help_text': _(
-                'The GitHub App ID created by the GitHub Admin. '
-                'Example App ID: 1121547 '
+                'The GitHub App ID or Client ID created by the GitHub Admin. '
+                'Example App ID: 1121547, Client ID: Iv23likIfIXeZTb5GCAA '
                 'found on https://github.com/settings/apps/ '
                 'required for creating a JWT token for authentication.',
             ),
@@ -130,65 +130,6 @@ class EmptyKwargs(TypedDict):
     """Schema for zero keyword arguments."""
 
 
-GH_CLIENT_ID_TRAILER_LENGTH = 16
-HEXADECIMAL_BASE = 16
-
-
-def _is_intish(app_id_candidate: str | int) -> bool:
-    return isinstance(app_id_candidate, int) or app_id_candidate.isdigit()
-
-
-def _is_client_id(client_id_candidate: str) -> bool:
-    client_id_prefix = 'Iv1.'
-    if not client_id_candidate.startswith(client_id_prefix):
-        return False
-
-    client_id_trailer = client_id_candidate.removeprefix(client_id_prefix)
-
-    if len(client_id_trailer) != GH_CLIENT_ID_TRAILER_LENGTH:
-        return False
-
-    try:
-        int(client_id_trailer, base=HEXADECIMAL_BASE)
-    except ValueError:
-        return False
-
-    return True
-
-
-def _is_app_or_client_id(app_or_client_id_candidate: str | int) -> bool:
-    if _is_intish(app_or_client_id_candidate):
-        return True
-
-    assert not isinstance(app_or_client_id_candidate, int)  # type narrowing
-    return _is_client_id(app_or_client_id_candidate)
-
-
-def _assert_ids_look_acceptable(
-    app_or_client_id: int | str,
-    install_id: int | str,
-) -> None:
-    if not _is_app_or_client_id(app_or_client_id):
-        raise ValueError(
-            'Expected GitHub App or Client ID to be an integer or a string '
-            f'starting with `Iv1.` followed by 16 hexadecimal digits, '
-            f'but got {app_or_client_id!r}',
-        )
-
-    if isinstance(app_or_client_id, str) and _is_client_id(app_or_client_id):
-        raise ValueError(
-            'Expected GitHub App ID must be an integer or a string '
-            f'with an all-digit value, but got {app_or_client_id!r}. '
-            'Client IDs are currently unsupported.',
-        )
-
-    if not _is_intish(install_id):
-        raise ValueError(
-            'Expected GitHub App Installation ID to be an integer'
-            f' but got {install_id!r}',
-        )
-
-
 def extract_github_app_install_token(  # noqa: WPS210
     *,
     github_api_url: str,
@@ -200,17 +141,15 @@ def extract_github_app_install_token(  # noqa: WPS210
     """Generate a GH App Installation access token.
 
     :param github_api_url: The GitHub instance API endpoint URL.
-    :param app_or_client_id: The GitHub App ID.
+    :param app_or_client_id: The GitHub App ID or Client ID.
     :param private_rsa_key: The private key associated with the GitHub
         App.
     :param install_id: The GitHub App Installation ID.
     :param _discarded_kwargs: Aren't expected to be passed.
     :returns: A GitHub access token for a GitHub App Installation.
-    :raises ValueError: If any required parameters are invalid.
-    :raises RuntimeError: If any required parameters are invalid.
+    :raises ValueError: If authentication fails or parameters are invalid.
+    :raises RuntimeError: If an unexpected error occurs during token generation.
     """
-    _assert_ids_look_acceptable(app_or_client_id, install_id)
-
     auth = Auth.AppAuth(
         app_id=str(app_or_client_id),
         private_key=private_rsa_key,
--- a/tests/github_app_test.py
+++ b/tests/github_app_test.py
@@ -29,14 +29,6 @@ from jwt import decode as decode_jwt
 from awx_plugins.credentials import github_app as gh_app_plugin_mod
 
 
-github_app_jwt_client_id_unsupported = pytest.mark.xfail(
-    raises=(AssertionError, ValueError),
-    reason='Client ID in JWT is not currently supported by '
-    'PyGitHub and is disabled.\n\n'
-    'Ref: https://github.com/PyGithub/PyGithub/issues/3213',
-)
-
-
 RSA_PUBLIC_EXPONENT = 65_537  # noqa: WPS303
 MINIMUM_RSA_KEY_SIZE = 1024  # the lowest value chosen for performance in tests
 
@@ -95,92 +87,6 @@ class AppInstallIds(TypedDict):
 
     app_or_client_id: str
     install_id: str
-
-
-@pytest.mark.parametrize(
-    ('extract_github_app_install_token_args', 'expected_error_msg'),
-    (
-        pytest.param(
-            {
-                'app_or_client_id': 'invalid',
-                'install_id': '666',
-            },
-            '^Expected GitHub App or Client ID to be an integer or a string '
-            r'starting with `Iv1\.` followed by 16 hexadecimal digits, but got'
-            " 'invalid'$",
-            id='gh-app-id-broken-text',
-        ),
-        pytest.param(
-            {
-                'app_or_client_id': 'Iv1.bbbbbbbbbbbbbbb',
-                'install_id': '666',
-            },
-            '^Expected GitHub App or Client ID to be an integer or a string '
-            r'starting with `Iv1\.` followed by 16 hexadecimal digits, but got'
-            " 'Iv1.bbbbbbbbbbbbbbb'$",
-            id='gh-app-id-client-id-not-enough-chars',
-        ),
-        pytest.param(
-            {
-                'app_or_client_id': 'Iv1.bbbbbbbbbbbbbbbx',
-                'install_id': '666',
-            },
-            '^Expected GitHub App or Client ID to be an integer or a string '
-            r'starting with `Iv1\.` followed by 16 hexadecimal digits, but got'
-            " 'Iv1.bbbbbbbbbbbbbbbx'$",
-            id='gh-app-id-client-id-broken-hex',
-        ),
-        pytest.param(
-            {
-                'app_or_client_id': 'Iv1.bbbbbbbbbbbbbbbbb',
-                'install_id': '666',
-            },
-            '^Expected GitHub App or Client ID to be an integer or a string '
-            r'starting with `Iv1\.` followed by 16 hexadecimal digits, but got'
-            " 'Iv1.bbbbbbbbbbbbbbbbb'$",
-            id='gh-app-id-client-id-too-many-chars',
-        ),
-        pytest.param(
-            {
-                'app_or_client_id': 999,
-                'install_id': 'invalid',
-            },
-            '^Expected GitHub App Installation ID to be an integer '
-            "but got 'invalid'$",
-            id='gh-app-invalid-install-id-with-int-app-id',
-        ),
-        pytest.param(
-            {
-                'app_or_client_id': '999',
-                'install_id': 'invalid',
-            },
-            '^Expected GitHub App Installation ID to be an integer '
-            "but got 'invalid'$",
-            id='gh-app-invalid-install-id-with-str-digit-app-id',
-        ),
-        pytest.param(
-            {
-                'app_or_client_id': 'Iv1.cccccccccccccccc',
-                'install_id': 'invalid',
-            },
-            '^Expected GitHub App Installation ID to be an integer '
-            "but got 'invalid'$",
-            id='gh-app-invalid-install-id-with-client-id',
-            marks=github_app_jwt_client_id_unsupported,
-        ),
-    ),
-)
-def test_github_app_invalid_args(
-    extract_github_app_install_token_args: AppInstallIds,
-    expected_error_msg: str,
-) -> None:
-    """Test that invalid arguments make token extractor bail early."""
-    with pytest.raises(ValueError, match=expected_error_msg):
-        gh_app_plugin_mod.extract_github_app_install_token(
-            github_api_url='https://github.com',
-            private_rsa_key='key',
-            **extract_github_app_install_token_args,
-        )
 
 
 @pytest.mark.parametrize(
@@ -271,16 +177,16 @@ class _FakeAppInstallationAuth(AppInstallationAuth):
 
 
 @pytest.mark.parametrize(
-    'application_id',
+    'application_or_client_id',
     (
         123,
         '123',
-        pytest.param(
-            'Iv1.aaaaaaaaaaaaaaaa',
-            marks=github_app_jwt_client_id_unsupported,
-        ),
+        'Iv1.aaaaaaaaaaaaaaaa',
+        'Iv2aaaaaaaaaaaaaaaa',
+        'Iv10aaaaaaaaaaaaaaaa',
+        'Iv100.aaaaaaaaaaaaaaaa',
+        'Iv23likIfIXeZTb5GCAA',
     ),
-    ids=('app-id-int', 'app-id-str', 'client-id'),
 )
 @pytest.mark.parametrize(
     'installation_id',
@@ -289,7 +195,7 @@ class _FakeAppInstallationAuth(AppInstallationAuth):
 )
 # pylint: disable-next=too-many-arguments,too-many-positional-arguments
 def test_github_app_github_authentication(  # noqa: WPS211
-    application_id: int | str,
+    application_or_client_id: int | str,
     installation_id: int | str,
     mocker: MockerFixture,
     monkeypatch: pytest.MonkeyPatch,
@@ -311,7 +217,7 @@ def test_github_app_github_authentication(  # noqa: WPS211
 
     token = gh_app_plugin_mod.extract_github_app_install_token(
         github_api_url='https://github.com',
-        app_or_client_id=application_id,
+        app_or_client_id=application_or_client_id,
         install_id=installation_id,
         private_rsa_key=rsa_private_key_str,
     )
@@ -346,6 +252,6 @@ def test_github_app_github_authentication(  # noqa: WPS211
             'verify_nbf': True,
         },
         audience=None,  # GH App JWT don't set the audience claim
-        issuer=str(application_id),
+        issuer=str(application_or_client_id),
         leeway=0.001,  # noqa: WPS432
     )
--- a/tests/github_app_test.py
+++ b/tests/github_app_test.py
@@ -31,6 +31,7 @@ from awx_plugins.credentials import github_app as gh_app_plugin_mod
 
 RSA_PUBLIC_EXPONENT = 65_537  # noqa: WPS303
 MINIMUM_RSA_KEY_SIZE = 1024  # the lowest value chosen for performance in tests
+TEST_APP_ID = 123
 
 
 @pytest.fixture(scope='module')
@@ -179,13 +180,13 @@ class _FakeAppInstallationAuth(AppInstallationAuth):
 @pytest.mark.parametrize(
     'application_or_client_id',
     (
-        123,
-        '123',
-        'Iv1.aaaaaaaaaaaaaaaa',
-        'Iv2aaaaaaaaaaaaaaaa',
-        'Iv10aaaaaaaaaaaaaaaa',
-        'Iv100.aaaaaaaaaaaaaaaa',
-        'Iv23likIfIXeZTb5GCAA',
+        pytest.param(TEST_APP_ID, id='app-id-int'),
+        pytest.param(str(TEST_APP_ID), id='app-id-str'),
+        pytest.param('Iv1.aaaaaaaaaaaaaaaa', id='client-id-iv1-legacy-format'),
+        pytest.param('Iv2aaaaaaaaaaaaaaaa', id='client-id-iv2-new-format'),
+        pytest.param('Iv10aaaaaaaaaaaaaaaa', id='client-id-iv10-double-digit-version'),
+        pytest.param('Iv100.aaaaaaaaaaaaaaaa', id='client-id-iv100-with-dot'),
+        pytest.param('Iv23likIfIXeZTb5GCAA', id='client-id-iv23-real-example'),
     ),
 )
 @pytest.mark.parametrize(
--- a/tests/github_app_test.py
+++ b/tests/github_app_test.py
@@ -184,7 +184,10 @@ class _FakeAppInstallationAuth(AppInstallationAuth):
         pytest.param(str(TEST_APP_ID), id='app-id-str'),
         pytest.param('Iv1.aaaaaaaaaaaaaaaa', id='client-id-iv1-legacy-format'),
         pytest.param('Iv2aaaaaaaaaaaaaaaa', id='client-id-iv2-new-format'),
-        pytest.param('Iv10aaaaaaaaaaaaaaaa', id='client-id-iv10-double-digit-version'),
+        pytest.param(
+            'Iv10aaaaaaaaaaaaaaaa',
+            id='client-id-iv10-double-digit-version',
+        ),
         pytest.param('Iv100.aaaaaaaaaaaaaaaa', id='client-id-iv100-with-dot'),
         pytest.param('Iv23likIfIXeZTb5GCAA', id='client-id-iv23-real-example'),
     ),


src/awx_plugins/credentials/github_app.py
```diff
@@ -139,8 +139,15 @@ def _is_intish(app_id_candidate: str | int) -> bool:
 
 
 def _is_client_id(client_id_candidate: str) -> bool:
-    client_id_prefix = 'Iv1.'
-    if not client_id_candidate.startswith(client_id_prefix):
+    # Support both Iv1 and Iv2 client ID prefixes
+    valid_prefixes = ('Iv1.', 'Iv2.')
```
arrestle: I would recommend allowing "Iv#." Since you don't want to break again if they come up with Iv3. next year.

arrestle (dismissed):
Great start! Python has https://docs.python.org/3/library/re.html or https://www.w3schools.com/python/python_regex.asp

src/awx_plugins/credentials/github_app.py
stevensonmichel: Sure

tests/github_app_test.py
```diff
@@ -279,8 +297,22 @@ def token(self: '_FakeAppInstallationAuth') -> str:
...
+            'Iv23likIfIXeZTb5GBUU',
+            marks=github_app_jwt_client_id_unsupported,
+        ),
+    ),
+    ids=(
+        'app-id-int',
+        'app-id-str',
+        'client-id-iv1',
+        'client-id-iv2',
+        'client-id-iv23',
```
webknjaz: These IDs don't explain what the respective tests are supposed to check. Could you make it clearer?

src/awx_plugins/credentials/github_app.py
```diff
@@ -132,28 +132,23 @@ class EmptyKwargs(TypedDict):
...
 
 def _is_client_id(client_id_candidate: str) -> bool:
-    client_id_prefix = 'Iv1.'
-    if not client_id_candidate.startswith(client_id_prefix):
-        return False
-
-    client_id_trailer = client_id_candidate.removeprefix(client_id_prefix)
-
-    if len(client_id_trailer) != GH_CLIENT_ID_TRAILER_LENGTH:
-        return False
```
webknjaz, line 149: It might be a good idea to have two validation paths — for v1 and for v2. Then, v1 could remain stricter.

tests/github_app_test.py
```diff
@@ -165,7 +173,17 @@ class AppInstallIds(TypedDict):
...
+        ),
+        pytest.param(
+            {
+                'app_or_client_id': 'Iv2.cccccccccccccccc',
+                'install_id': 'invalid',
+            },
+            '^Expected GitHub App Installation ID to be an integer '
+            "but got 'invalid'$",
+            id='gh-app-invalid-install-id-with-client-id-iv2',
             marks=github_app_jwt_client_id_unsupported,
```
webknjaz: Is this even needed given that the test checks the install ID and doesn't care what's in the app/client ID? The additiona seems redundant.

tests/github_app_test.py
```diff
@@ -165,7 +173,17 @@ class AppInstallIds(TypedDict):
             },
             '^Expected GitHub App Installation ID to be an integer '
             "but got 'invalid'$",
-            id='gh-app-invalid-install-id-with-client-id',
+            id='gh-app-invalid-install-id-with-client-id-iv1',
+            marks=github_app_jwt_client_id_unsupported,
```
webknjaz: Following the linked issue chain, it seems that it might work now. It would be worth checking if anything changed before makring another test with xfail.

tests/github_app_test.py
webknjaz: Additionally, with the list of tests becoming longer, it's probably a good idea to move IDs into `pytest.param()` definitions and keep them coupled. It's okay for 2-3 tests but it'd difficult to maintain two coupled iterables when they are long.

webknjaz:
@stevensonmichel FYI adding merge commits into PR branches creates foxtrots, which is why using rebase is preferred.

src/awx_plugins/credentials/github_app.py
```diff
@@ -156,6 +158,31 @@ def _is_client_id(client_id_candidate: str) -> bool:
     return True
 
 
+def _is_iv1_prefix(client_id_candidate: str) -> bool:
+    """Check if it starts with Iv1 (not Iv10, Iv100, etc.)."""
+    if not client_id_candidate.startswith('Iv1'):
+        return False
+    if len(client_id_candidate) <= 3:
+        return True
```
webknjaz: Why is this true?

src/awx_plugins/credentials/github_app.py
```diff
@@ -132,18 +132,20 @@ class EmptyKwargs(TypedDict):
 
 GH_CLIENT_ID_TRAILER_LENGTH = 16
 HEXADECIMAL_BASE = 16
+GH_CLIENT_ID_MIN_LENGTH = 19
+GH_CLIENT_ID_MAX_LENGTH = 22
```
webknjaz: Let's try keeping coupled values together:
```suggestion
GH_CLIENT_ID_LENGTH_BOUNDS = range(19, 23)  # min 19, max 22
```

src/awx_plugins/credentials/github_app.py
```diff
@@ -156,6 +158,31 @@ def _is_client_id(client_id_candidate: str) -> bool:
...
+
+    if _is_iv1_prefix(client_id_candidate):
+        return _is_iv1_client_id(client_id_candidate)
+
+    # For Iv2+, just check length
+    return (
+        client_id_candidate.startswith('Iv')
+        and GH_CLIENT_ID_MIN_LENGTH
+        <= len(client_id_candidate)
+        <= GH_CLIENT_ID_MAX_LENGTH
```
webknjaz:
```suggestion
        and len(client_id_candidate) in GH_CLIENT_ID_LENGTH_BOUNDS
```

webknjaz (commented):
I feel like keeping a range in the same object is a bit more readable.

src/awx_plugins/credentials/github_app.py
stevensonmichel: Alright! Resolved

src/awx_plugins/credentials/github_app.py
stevensonmichel: This is just to prevent and indexError crash. A suitable example would be a client id with only 3 characters

src/awx_plugins/credentials/github_app.py
stevensonmichel: Fixed

src/awx_plugins/credentials/github_app.py
stevensonmichel: Fixed

tests/github_app_test.py
stevensonmichel: Fixed

src/awx_plugins/credentials/github_app.py
webknjaz: Turned out this is confusing to parse because past the first guard expression, all values of `client_id_candidate` would start with `Iv1` which is 3 exactly. And checking for less than 3 is counterintuitive.

src/awx_plugins/credentials/github_app.py
```diff
@@ -156,6 +157,29 @@ def _is_client_id(client_id_candidate: str) -> bool:
     return True
 
 
+def _is_iv1_prefix(client_id_candidate: str) -> bool:
+    """Check if it starts with Iv1 (not Iv10, Iv100, etc.)."""
+    if not client_id_candidate.startswith('Iv1'):
+        return False
+    if len(client_id_candidate) <= 3:
+        return True
+    return not client_id_candidate[3].isdigit()
```
webknjaz:
```suggestion
    known_v1_prefix = 'Iv1'

    if not client_id_candidate.startswith(known_v1_prefix):
        return False

    client_id_candidate_trailer = client_id_candidate.removeprefix(known_v1_prefix)

    if not client_id_candidate_trailer:  # empty trailer, no digits to check
        return True

    return not client_id_candidate_trailer[0].isdigit()
```

src/awx_plugins/credentials/github_app.py
```diff
@@ -156,6 +157,29 @@ def _is_client_id(client_id_candidate: str) -> bool:
     return True
 
 
+def _is_iv1_prefix(client_id_candidate: str) -> bool:
```
webknjaz: Looking closer, I don't understand why you don't check for `Iv1.`

src/awx_plugins/credentials/github_app.py
```diff
@@ -156,6 +157,29 @@ def _is_client_id(client_id_candidate: str) -> bool:
...
+        return True
+    return not client_id_candidate[3].isdigit()
+
+
+def _is_client_id(client_id_candidate: str) -> bool:
+    # Iv1: strict format with Iv1. + 16 hex chars
+    # Iv2+: just check length (19-22 chars)
+
+    if _is_iv1_prefix(client_id_candidate):
+        return _is_iv1_client_id(client_id_candidate)
```
webknjaz: Why can't this just be
```suggestion
    if _is_iv1_client_id(client_id_candidate):
        return True
```
?

src/awx_plugins/credentials/github_app.py
```diff
@@ -132,18 +132,19 @@ class EmptyKwargs(TypedDict):
...
-def _is_client_id(client_id_candidate: str) -> bool:
-    client_id_prefix = 'Iv1.'
-    if not client_id_candidate.startswith(client_id_prefix):
+def _is_iv1_client_id(client_id_candidate: str) -> bool:
+    """Check if it's a valid Iv1 client ID (strict: Iv1. + 16 hex chars)."""
+    if not client_id_candidate.startswith('Iv1.'):
         return False
 
-    client_id_trailer = client_id_candidate.removeprefix(client_id_prefix)
+    client_id_trailer = client_id_candidate.removeprefix('Iv1.')
```
webknjaz: No need to duplicate literals in multiple places. Revert this to the original forms.
```suggestion
    client_id_prefix = 'Iv1.'
    if not client_id_candidate.startswith(client_id_prefix):
        return False

    client_id_trailer = client_id_candidate.removeprefix(client_id_prefix)
```

src/awx_plugins/credentials/github_app.py
```diff
@@ -156,6 +157,29 @@ def _is_client_id(client_id_candidate: str) -> bool:
...
 
+def _is_iv1_prefix(client_id_candidate: str) -> bool:
+    """Check if it starts with Iv1 (not Iv10, Iv100, etc.)."""
+    if not client_id_candidate.startswith('Iv1'):
+        return False
+    if len(client_id_candidate) <= 3:
+        return True
+    return not client_id_candidate[3].isdigit()
+
+
```
webknjaz: With other observations, this becomes unnecessary.
```suggestion
```

src/awx_plugins/credentials/github_app.py
```diff
@@ -132,18 +132,19 @@ class EmptyKwargs(TypedDict):
 
 GH_CLIENT_ID_TRAILER_LENGTH = 16
 HEXADECIMAL_BASE = 16
+GH_CLIENT_ID_LENGTH_BOUNDS = (19, 22)
```
webknjaz: Let's unbreak this by applying the original suggestion I made in https://github.com/ansible/awx-plugins/pull/139#discussion_r2473386861
```suggestion
GH_CLIENT_ID_LENGTH_BOUNDS = range(19, 23)  # min 19, max 22
```

`thing in (1, 42)` will only be true if `thing == 1 or thing == 42`. Using the range allows anything in-between. Just take into account that the second argument is not included in the range. So `3 in range(1, 42) is True` and `41 in range(1, 42) is True`, but `42 in range(1, 42) is False`.

tests/github_app_test.py
```diff
@@ -106,39 +108,100 @@ class AppInstallIds(TypedDict):
...
+        pytest.param(
+            {
+                'app_or_client_id': 'Iv23toolongtobevalidids',
+                'install_id': '666',
+            },
+            '^Expected GitHub App or Client ID to be an integer or a string '
+            r'starting with `Iv1\.` \(16 hex chars\) or `Iv2\+` \(19-22 chars\), '
+            'but got'
+            " 'Iv23toolongtobevalidids'$",
+            id='gh-app-id-iv2-too-long',
```
webknjaz: To catch bugs like the one I found in, https://github.com/ansible/awx-plugins/pull/139/files?new_files_changed=true#r2478577419, add tests for the lengths matching the boundaries — with 18/19/22/23 chars exactly.

tests/github_app_test.py
webknjaz: @stevensonmichel I'm still waiting for the results. Marking the thread as resolved does not mean that this is resolved. It's not.

tests/github_app_test.py
```diff
@@ -273,19 +346,36 @@ def token(self: '_FakeAppInstallationAuth') -> str:
...
     ),
-    ids=('app-id-int', 'app-id-str', 'client-id'),
 )
 @pytest.mark.parametrize(
     'installation_id',
-    (456, '456'),
-    ids=('install-id-int', 'install-id-str'),
+    (
+        pytest.param(TEST_INSTALLATION_ID, id='install-id-int'),
+        pytest.param(str(TEST_INSTALLATION_ID), id='install-id-str-numeric'),
```
webknjaz: I wouldn't touch unrelated places. This will end up extending the scope of the PR infinitely. It's a bad tone to smash together different types of changes (like refactoring/formatting/functional). Pick one and stick with it so that the PR is atomic and only contains changes that contribute to a single declared goal.

djyasin (approved):

src/awx_plugins/credentials/github_app.py
arrestle: because github switched to access tokens with a prefix of Iv2 with possible extensions to other prefixes... I'm not sure checking for ANY prefix makes sense any more. see https://docs.github.com/en/rest/apps/oauth-applications?apiVersion=2022-11-28

tests/github_app_test.py
arrestle: I would remove Iv1 and Iv2 - possibly just `^Iv[1-9]\d*\.[a-zA-Z0-9]+$` or use the endpoint 

`curl -L \
  -X POST \
  -H "Accept: application/vnd.github+json" \
  -u "<YOUR_CLIENT_ID>:<YOUR_CLIENT_SECRET>" \
  -H "X-GitHub-Api-Version: 2022-11-28" \
  "https://api.github.com/applications/<CLIENT_ID>/token" \
  -d '{"access_token":"<TOKEN_TO_VALIDATE>"}'`
  
See https://docs.github.com/en/rest/apps/oauth-applications?apiVersion=2022-11-28#check-a-token or https://docs.github.com/en/rest/apps/oauth-applications?apiVersion=2022-11-28

webknjaz: Well, since there's already network I/O for retrieving the token, it will tell us if the creds are valid right in there. But it probably won't specify whether it's the client id or the token that's wrong.
If a regex was to be used, it'd probably need to differentiate between v1 and the rest. Though, I'm not a fan of regexes for such simple checks since they're more difficult to maintain. The coverage report won't tell you that you messed us the range boundaries in your regex, for example and it won't hint you how to cover unobvious cases.

src/awx_plugins/credentials/github_app.py
webknjaz: Not sure what that doc demonstrates. We already know that any validation is hunch-based. I'm actually starting to think that validation would be more useful in a form of a non-blocking hint rather than a thing that tries to predict whether input is valid. At least, for the inputs where GH doesn't document their exact shape.

src/awx_plugins/credentials/github_app.py
```diff
@@ -156,6 +157,28 @@ def _is_client_id(client_id_candidate: str) -> bool:
...
 
 
+def _is_iv1_prefix(client_id_candidate: str) -> bool:
+    """Check if it starts with Iv1 (not Iv10, Iv100, etc.)."""
+    known_v1_prefix = 'Iv1'
+
+    if not client_id_candidate.startswith(known_v1_prefix):
+        return False
+
+    client_id_candidate_trailer = client_id_candidate.removeprefix(
```
webknjaz: This should unbreak the CI
```suggestion
```

(*I think* it's GitHub's new experimental review interface that might've corrupted the patch)

src/awx_plugins/credentials/github_app.py
```diff
@@ -132,18 +132,19 @@ class EmptyKwargs(TypedDict):
...
-def _is_client_id(client_id_candidate: str) -> bool:
-    client_id_prefix = 'Iv1.'
-    if not client_id_candidate.startswith(client_id_prefix):
+def _is_iv1_client_id(client_id_candidate: str) -> bool:
+    """Check if it's a valid Iv1 client ID (strict: Iv1. + 16 hex chars)."""
+    if not client_id_candidate.startswith('Iv1.'):
         return False
 
-    client_id_trailer = client_id_candidate.removeprefix(client_id_prefix)
+    client_id_trailer = client_id_candidate.removeprefix('Iv1.')
```
webknjaz:
```suggestion
    client_id_prefix = 'Iv1.'
    if not client_id_candidate.startswith(client_id_prefix):
        return False

    client_id_trailer = client_id_candidate.removeprefix(client_id_prefix)
```

tests/github_app_test.py
```diff
@@ -106,39 +108,100 @@ class AppInstallIds(TypedDict):
...
+                'app_or_client_id': 'Iv1',
                 'install_id': '666',
             },
             '^Expected GitHub App or Client ID to be an integer or a string '
-            r'starting with `Iv1\.` followed by 16 hexadecimal digits, but got'
-            " 'Iv1.bbbbbbbbbbbbbbb'$",
-            id='gh-app-id-client-id-not-enough-chars',
+            r'starting with `Iv1\.` \(16 hex chars\) or `Iv2\+` \(19-22 chars\), '
+            'but got'
+            " 'Iv1'$",
```
webknjaz:
```suggestion
            "but got 'Iv1'$",
```

tests/github_app_test.py
```diff
@@ -106,39 +108,100 @@ class AppInstallIds(TypedDict):
...
+                'app_or_client_id': 'Iv1.bbb',
                 'install_id': '666',
             },
             '^Expected GitHub App or Client ID to be an integer or a string '
-            r'starting with `Iv1\.` followed by 16 hexadecimal digits, but got'
-            " 'Iv1.bbbbbbbbbbbbbbbx'$",
-            id='gh-app-id-client-id-broken-hex',
+            r'starting with `Iv1\.` \(16 hex chars\) or `Iv2\+` \(19-22 chars\), '
+            'but got'
+            " 'Iv1.bbb'$",
```
webknjaz:
```suggestion
            "but got 'Iv1.bbb'$",
```

tests/github_app_test.py
```diff
@@ -106,39 +108,100 @@ class AppInstallIds(TypedDict):
...
+                'app_or_client_id': 'Iv1aaaaaaaaaaaaaaaa',
                 'install_id': '666',
             },
             '^Expected GitHub App or Client ID to be an integer or a string '
-            r'starting with `Iv1\.` followed by 16 hexadecimal digits, but got'
-            " 'Iv1.bbbbbbbbbbbbbbbbb'$",
-            id='gh-app-id-client-id-too-many-chars',
+            r'starting with `Iv1\.` \(16 hex chars\) or `Iv2\+` \(19-22 chars\), '
+            'but got'
+            " 'Iv1aaaaaaaaaaaaaaaa'$",
```
webknjaz:
```suggestion
            "but got 'Iv1aaaaaaaaaaaaaaaa'$",
```

tests/github_app_test.py
```diff
@@ -106,39 +108,100 @@ class AppInstallIds(TypedDict):
...
+        ),
+        pytest.param(
+            {
+                'app_or_client_id': 'Iv1.gggggggggggggggg',
+                'install_id': '666',
+            },
+            '^Expected GitHub App or Client ID to be an integer or a string '
+            r'starting with `Iv1\.` \(16 hex chars\) or `Iv2\+` \(19-22 chars\), '
+            'but got'
+            " 'Iv1.gggggggggggggggg'$",
```
webknjaz:
```suggestion
            "but got 'Iv1.gggggggggggggggg'$",
```

tests/github_app_test.py
```diff
@@ -106,39 +108,100 @@ class AppInstallIds(TypedDict):
...
+        ),
+        pytest.param(
+            {
+                'app_or_client_id': 'Iv1.aaaaaaaaaaaaaaaaa',
+                'install_id': '666',
+            },
+            '^Expected GitHub App or Client ID to be an integer or a string '
+            r'starting with `Iv1\.` \(16 hex chars\) or `Iv2\+` \(19-22 chars\), '
+            'but got'
+            " 'Iv1.aaaaaaaaaaaaaaaaa'$",
```
webknjaz:
```suggestion
            "but got 'Iv1.aaaaaaaaaaaaaaaaa'$",
```

tests/github_app_test.py
```diff
@@ -106,39 +108,100 @@ class AppInstallIds(TypedDict):
...
+        # Iv2+ validation tests (relaxed: just length)
+        pytest.param(
+            {
+                'app_or_client_id': 'Iv',
+                'install_id': '666',
+            },
+            '^Expected GitHub App or Client ID to be an integer or a string '
+            r'starting with `Iv1\.` \(16 hex chars\) or `Iv2\+` \(19-22 chars\), '
+            'but got'
+            " 'Iv'$",
```
webknjaz:
```suggestion
            "but got 'Iv'$",
```

tests/github_app_test.py
```diff
@@ -106,39 +108,100 @@ class AppInstallIds(TypedDict):
...
+        ),
+        pytest.param(
+            {
+                'app_or_client_id': 'Iv2.bbb',
+                'install_id': '666',
+            },
+            '^Expected GitHub App or Client ID to be an integer or a string '
+            r'starting with `Iv1\.` \(16 hex chars\) or `Iv2\+` \(19-22 chars\), '
+            'but got'
+            " 'Iv2.bbb'$",
```
webknjaz:
```suggestion
            "but got 'Iv2.bbb'$",
```

tests/github_app_test.py
```diff
@@ -106,39 +108,100 @@ class AppInstallIds(TypedDict):
...
+        ),
+        pytest.param(
+            {
+                'app_or_client_id': 'Iv23toolongtobevalidids',
+                'install_id': '666',
+            },
+            '^Expected GitHub App or Client ID to be an integer or a string '
+            r'starting with `Iv1\.` \(16 hex chars\) or `Iv2\+` \(19-22 chars\), '
+            'but got'
+            " 'Iv23toolongtobevalidids'$",
```
webknjaz:
```suggestion
            "but got 'Iv23toolongtobevalidids'$",
```

arrestle (dismissed):
I would avoid relying on string comparison and use [this github endpoint](https://docs.github.com/en/rest/authentication/authenticating-to-the-rest-api?apiVersion=2022-11-28) to validate a token, if indeed this is even a requirement. 

The github token code may be in flux, and while currently seems to be Iv[1|2].[n+] future versions may not even follow that pattern.

src/awx_plugins/credentials/github_app.py
```diff
@@ -132,28 +133,48 @@ class EmptyKwargs(TypedDict):
 
 GH_CLIENT_ID_TRAILER_LENGTH = 16
 HEXADECIMAL_BASE = 16
+GH_CLIENT_ID_MIN_LENGTH = 19
+GH_CLIENT_ID_MAX_LENGTH = 22
+GH_CLIENT_ID_LENGTH_BOUNDS = range(
+    GH_CLIENT_ID_MIN_LENGTH,
+    GH_CLIENT_ID_MAX_LENGTH + 1,
+)  # min 19, max 22
```
webknjaz: Since you've reintroduced explicit constants, this comment is no longer necessary. In fact, there is harm in it potentially going out of sync and becoming confusing.
```suggestion
)
```

src/awx_plugins/credentials/github_app.py
```diff
@@ -132,28 +133,48 @@ class EmptyKwargs(TypedDict):
...
 
 
 def _is_intish(app_id_candidate: str | int) -> bool:
     return isinstance(app_id_candidate, int) or app_id_candidate.isdigit()
 
 
 def _is_client_id(client_id_candidate: str) -> bool:
```
webknjaz: Plz use a helper function instead of increasing indentation.

src/awx_plugins/credentials/github_app.py
```diff
@@ -156,6 +163,28 @@ def _is_client_id(client_id_candidate: str) -> bool:
...
+    # Iv1 + non-digit (Iv1aaa): INVALID
+    # Iv2+: just check length (19-22 chars)
+
+    # Check for Iv1. with dot - strict validation
+    if _is_iv1_client_id(client_id_candidate):
+        return True
+
+    # Check for invalid Iv1 format: Iv1 followed by non-digit
+    # (not dot, not digit). Examples: Iv1aaa, Iv1xxx → INVALID
+    if re.match(r'^Iv1[^\d]', client_id_candidate):
```
webknjaz: At this point, it should never be a v1 since `_is_iv1_client_id()` would be `False` here. But if we want this additional check, it belongs in that helper and shouldn't be in disconnected context.

webknjaz:
In general, I agree with @arrestle in that we don't know anything about these tokens and might as well just drop any shape/length validation.

P.S. I'm still interested in learning if the underlying libraries are finally able to accept client IDs, though, and dropping `xfails` so it's actually tested somehow.

src/awx_plugins/credentials/github_app.py
stevensonmichel: Now, I see. As it is to validate cases like Iv10.... , I think my comment "Check for Iv1 format" is misleading. I will change that.

stevensonmichel: Also, I tested PyGithub with client Ids, and it validated it return the token ghs_...

tests/github_app_test.py
```diff
@@ -97,90 +93,9 @@ class AppInstallIds(TypedDict):
...
-    """Test that invalid arguments make token extractor bail early."""
-    with pytest.raises(ValueError, match=expected_error_msg):
-        gh_app_plugin_mod.extract_github_app_install_token(
-            github_api_url='https://github.com',
-            private_rsa_key='key',
-            **extract_github_app_install_token_args,
-        )
+# NOTE: Format validation tests removed - we no longer validate
+# Client ID formats upfront. GitHub's API is the ultimate validator.
+# Invalid credentials will be rejected by the API with appropriate errors.
```
webknjaz: Make sure not to keep comments that aren't useful in files.

tests/github_app_test.py
```diff
@@ -29,12 +29,8 @@
...
 
 
-github_app_jwt_client_id_unsupported = pytest.mark.xfail(
-    raises=(AssertionError, ValueError),
-    reason='Client ID in JWT is not currently supported by '
-    'PyGitHub and is disabled.\n\n'
-    'Ref: https://github.com/PyGithub/PyGithub/issues/3213',
-)
+# Client IDs are now supported - no longer using xfail marker
+# All Client ID tests should pass with the no-validation approach
```
webknjaz: This detached statement is not needed. It doesn't add clarity, just confusion.

tests/github_app_test.py
webknjaz: rename this to
```suggestion
    'application_or_client_id',
```

webknjaz:
@stevensonmichel macos will stop failing once you rebase.

tests/github_app_test.py
```diff
@@ -271,16 +177,16 @@ def token(self: '_FakeAppInstallationAuth') -> str:
...
         '123',
-        pytest.param(
-            'Iv1.aaaaaaaaaaaaaaaa',
-            marks=github_app_jwt_client_id_unsupported,
-        ),
+        'Iv1.aaaaaaaaaaaaaaaa',
+        'Iv2aaaaaaaaaaaaaaaa',
+        'Iv10aaaaaaaaaaaaaaaa',
+        'Iv100.aaaaaaaaaaaaaaaa',
+        'Iv23likIfIXeZTb5GCAA',
```
webknjaz: These params become a part of the test ID. They show up in test reports and console output. It's not necessarily clear what they are for.
This is why, please wrap each entry with `pytest.param(..., id='human-oriented-test-case-identifier')` which would make it clear why these are being tested in the first place.
Don't use the separate `ids=` argument to `pytest.mark.parametrize()` since the list of things is too long for that style to be maintainable in this specific case.

src/awx_plugins/credentials/github_app.py
webknjaz: > I tested PyGithub with client Ids, and it validated it return the token ghs_...

@stevensonmichel good to know. I see that you've already taken care of the `xfail` markers so I'm marking this thread as resolved.

tests/github_app_test.py
webknjaz: Marking this thread as resolved.

tests/github_app_test.py
webknjaz: Marking this thread as resolved.

tests/github_app_test.py
webknjaz: Marking this thread as resolved.

src/awx_plugins/credentials/github_app.py
webknjaz: Marking this thread as resolved.

tests/github_app_test.py
webknjaz: Marking this thread as resolved.

src/awx_plugins/credentials/github_app.py
webknjaz: Marking this thread as resolved.

src/awx_plugins/credentials/github_app.py
webknjaz: Marking this thread as resolved.

src/awx_plugins/credentials/github_app.py
webknjaz: Marking this thread as resolved.

src/awx_plugins/credentials/github_app.py
webknjaz: Marking this thread as resolved.

src/awx_plugins/credentials/github_app.py
webknjaz: Marking this thread as resolved.

src/awx_plugins/credentials/github_app.py
webknjaz: Marking this thread as resolved.

arrestle (approved):
Hi Stevenson! Good job getting this reviewed it's a tough one. 

I approve of the approach but honestly don't completely understand the implications of changing application_id to  application_or_client_id. However, I trust your team has enough integration testing to make sure this all works correctly! 

Let me know if you need any further assistance.

stevensonmichel:
Thank you so much for the review and approval! I really appreciate your feedback. Regarding the application_id to application_or_client_id change, the name was updated to better reflect that it now accepts both GitHub App IDs and Client IDs. The key change in this PR is letting PyGitHub's API be the ultimate validator. This makes the plugin future-proof when GitHub introduces new ID formats.

PR #140, opened 2025-11-01
Add a managed credential type for HCP Terraform

shvenkat-rh:
The plugins in the hashicorp.terraform collection requires a token to authenticate with TFE/C. When running playbooks leveraging these modules on AAP, this token either needs to be set in the playbook/tasks or a custom credential type needs to be created that injects the token, via the env variable TF_TOKEN. This should be a built-in credential type in Controller for a better UX. There's a credential type already in AAP for Terraform. But that only deals with the backend configuration section which is not valid for TFE/C. 

This PR adds support for the HCP Credential.

https://issues.redhat.com/browse/AAP-57221

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -101,6 +101,7 @@ insights = "awx_plugins.credentials.plugins:insights"
 rhv = "awx_plugins.credentials.plugins:rhv"
 gpg_public_key = "awx_plugins.credentials.plugins:gpg_public_key"
 terraform = "awx_plugins.credentials.plugins:terraform"
+hcp_terraform = "awx_plugins.credentials._managed_types.terraform:hcp_terraform"
 
 [project.entry-points."awx_plugins.inventory"]  # new entry points group name
 azure_rm = "awx_plugins.inventory.plugins:azure_rm"
--- a/src/awx_plugins/credentials/_managed_types/terraform.py
+++ b/src/awx_plugins/credentials/_managed_types/terraform.py
@@ -1,0 +1,46 @@
+"""HCP Terraform managed credential type."""
+
+from awx_plugins.interfaces._temporary_private_api import (  # noqa: WPS436
+    ManagedCredentialType,
+)
+from awx_plugins.interfaces._temporary_private_django_api import (  # noqa: WPS436
+    gettext_noop,
+)
+
+from ..injectors import hcp_terraform as hcp_terraform_injector
+
+
+# FIXME: ManagedCredentialType and gettext_noop return Any from
+# _temporary_private_api interfaces. These nested Any structures cascade
+# into making outer dicts/lists typed as Any as well, requiring suppressions
+# throughout. Remove once those interfaces have proper type annotations.
+hcp_terraform = ManagedCredentialType(  # type: ignore[misc]
+    namespace='hcp_terraform',
+    kind='cloud',
+    name=gettext_noop('HCP Terraform'),  # type: ignore[misc]
+    managed=True,
+    custom_injectors=hcp_terraform_injector,
+    inputs={
+        'fields': [  # type: ignore[misc]
+            {  # type: ignore[misc]
+                'id': 'hostname',
+                'label': gettext_noop('Hostname'),  # type: ignore[misc]
+                'type': 'string',
+                'help_text': gettext_noop(  # type: ignore[misc]
+                    'The hostname of your HCP Terraform instance (e.g., app.terraform.io)',
+                ),
+                'default': 'app.terraform.io',
+            },
+            {  # type: ignore[misc]
+                'id': 'token',
+                'label': gettext_noop('API Token'),  # type: ignore[misc]
+                'type': 'string',
+                'secret': True,
+                'help_text': gettext_noop(  # type: ignore[misc]
+                    'HCP Terraform API Token',
+                ),
+            },
+        ],
+        'required': ['token'],
+    },
+)
--- a/src/awx_plugins/credentials/injectors.py
+++ b/src/awx_plugins/credentials/injectors.py
@@ -231,3 +231,14 @@ def terraform(
             path,
             private_data_dir,
         )
+
+
+def hcp_terraform(
+    cred: Credential,
+    env: EnvVarsType,
+    private_data_dir: str,
+) -> None:
+    env['TF_TOKEN'] = str(cred.get_input('token', default=''))
+    env['TF_HOSTNAME'] = str(
+        cred.get_input('hostname', default='app.terraform.io'),
+    )
--- a/tests/importable_test.py
+++ b/tests/importable_test.py
@@ -73,6 +73,15 @@ credential_plugins = (
 )
 
 
+managed_credential_plugins = (
+    EntryPointParam(
+        'awx_plugins.managed_credentials.supported',
+        'hcp_terraform',
+        'awx_plugins.credentials._managed_types.terraform:hcp_terraform',
+    ),
+)
+
+
 inventory_plugins = (
     EntryPointParam(
         'awx_plugins.inventory',
@@ -174,6 +183,13 @@ with_credential_plugins = pytest.mark.parametrize(
 )
 
 
+with_managed_credential_plugins = pytest.mark.parametrize(
+    'entry_point',
+    managed_credential_plugins,
+    ids=str,
+)
+
+
 with_inventory_plugins = pytest.mark.parametrize(
     'entry_point',
     inventory_plugins,
@@ -183,7 +199,7 @@ with_inventory_plugins = pytest.mark.parametrize(
 
 with_all_plugins = pytest.mark.parametrize(
     'entry_point',
-    credential_plugins + inventory_plugins,
+    credential_plugins + managed_credential_plugins + inventory_plugins,
     ids=str,
 )
 
@@ -211,6 +227,18 @@ def test_entry_points_are_credential_plugin(
 
     loaded_plugin_class_name = type(loaded_plugin_class).__name__
     assert loaded_plugin_class_name == 'CredentialPlugin'
+
+
+@with_managed_credential_plugins
+def test_entry_points_are_managed_credential_type(
+    entry_point: EntryPointParam,
+) -> None:
+    """Ensure all exposed managed credential plugins are of the same class."""  # noqa: D200, DAR101; FIXME
+    entry_points = _discover_entry_points(group=entry_point.group)
+    loaded_plugin_class = entry_points[entry_point.name].load()
+
+    loaded_plugin_class_name = type(loaded_plugin_class).__name__
+    assert loaded_plugin_class_name == 'ManagedCredentialType'
 
 
 @with_inventory_plugins
--- a/tests/managed_credential_plugins_test.py
+++ b/tests/managed_credential_plugins_test.py
@@ -28,6 +28,7 @@ from awx_plugins.interfaces._temporary_private_inject_api import (  # noqa: WPS4
 
 import yaml
 
+from awx_plugins.credentials._managed_types.terraform import hcp_terraform
 from awx_plugins.credentials.plugins import (  # noqa: WPS235
     aws,
     azure_rm,
@@ -454,6 +455,47 @@ class YamlEnvFile(BaseEnvFile[dict[str, IniEntryDataType]]):
                 JsonEnvFile('GOOGLE_BACKEND_CREDENTIALS', get_gce_creds()),
             ],
             id='terraform-gcs',
+        ),
+        pytest.param(
+            hcp_terraform,
+            {
+                'token': 'hcp-token-123',
+            },
+            {
+                'TF_TOKEN': 'hcp-token-123',
+                'TF_HOSTNAME': 'app.terraform.io',
+            },
+            {'TF_TOKEN': HIDDEN_PASSWORD},
+            [],
+            id='hcp-terraform-default-hostname',
+        ),
+        pytest.param(
+            hcp_terraform,
+            {
+                'hostname': 'terraform.example.com',
+                'token': 'hcp-custom-token-456',
+            },
+            {
+                'TF_TOKEN': 'hcp-custom-token-456',
+                'TF_HOSTNAME': 'terraform.example.com',
+            },
+            {'TF_TOKEN': HIDDEN_PASSWORD},
+            [],
+            id='hcp-terraform-custom-hostname',
+        ),
+        pytest.param(
+            hcp_terraform,
+            {
+                'hostname': 'app.terraform.io',
+                'token': 'hcp-explicit-default-789',
+            },
+            {
+                'TF_TOKEN': 'hcp-explicit-default-789',
+                'TF_HOSTNAME': 'app.terraform.io',
+            },
+            {'TF_TOKEN': HIDDEN_PASSWORD},
+            [],
+            id='hcp-terraform-explicit-default-hostname',
         ),
         pytest.param(
             net,


webknjaz:
You should probably add some entries to https://github.com/ansible/awx-plugins/blob/devel/tests/importable_test.py. Also, rebase the PR since there's macOS fixes applied due to GH deprecating older macOS VMs.

shvenkat-rh:
@webknjaz do we need to add something to  https://github.com/ansible/awx-plugins/blob/devel/tests/importable_test.py   for HCP Terraform credential even though it is a  ManagedCredentialType? I dont see any  ManagedCredentialType listed here.

webknjaz:
@shvenkat-rh merge commits in the PR branch will cause ugly foxtrots in the Git tree. Could you rebase instead?

webknjaz:
> do we need to add something to https://github.com/ansible/awx-plugins/blob/devel/tests/importable_test.py for HCP Terraform credential even though it is a ManagedCredentialType? I dont see any ManagedCredentialType listed here.

@shvenkat-rh since you added entry points, it's a good idea to make sure they are actually usable through the standard plugin loading interface. That test module performs such tests. I looked into it and you're right, there's no tests for managed credentials. However, you should be able to add another test by example.

pyproject.toml
```diff
@@ -86,6 +86,7 @@ kubernetes_bearer_token = "awx_plugins.credentials.plugins:kubernetes_bearer_tok
 registry = "awx_plugins.credentials.plugins:registry"
 galaxy_api_token = "awx_plugins.credentials.plugins:galaxy_api_token"
 net = "awx_plugins.credentials.plugins:net"
+hcp_terraform = "awx_plugins.credentials.plugins:hcp_terraform"
```
webknjaz: I think the exposed name should avoid abbreviations. It's often difficult to track what they mean, especially for people outside the context.
How about
```suggestion
hashicorp-terraform = "awx_plugins.credentials.plugins:hcp_terraform"
```
for example?

webknjaz: Ah, never mind. I google it and this seems to be a part of the official name. Marking the thread as resolved.

webknjaz: Note that this section contains a list of plugins loaded automatically, while the other one is loaded in supported variants of controller. I imagine the team might want to only keep the supported one down below, and drop this.

src/awx_plugins/credentials/plugins.py
webknjaz, line 1: This file is getting to big. What if instead of disabling linting, we just placed the new thing into a `src/awx_plugins/credentials/_managed_types/terraform.py`, keeping `src/awx_plugins/credentials/plugins.py` from growing infinitely?

src/awx_plugins/credentials/_managed_types/__init__.py
webknjaz, line 1: Don't add empty init modules. Keep this an implicit namespace.

webknjaz:
@shvenkat-rh you may want to squash the related commits at some point. Merge queues will make each of those a part of `devel` if you don't.

src/awx_plugins/credentials/plugins.py
```diff
@@ -1,5 +1,6 @@
 # FIXME: the following violations must be addressed gradually and unignored
 # mypy: disable-error-code="assignment, misc, no-redef"
+# pylint: disable=too-many-lines
```
webknjaz: This is redundant now.
```suggestion
```

src/awx_plugins/credentials/_managed_types/terraform.py
```diff
@@ -0,0 +1,45 @@
...
+# mypy: disable-error-code="assignment, misc"
+
+from awx_plugins.interfaces._temporary_private_api import (  # noqa: WPS436
+    ManagedCredentialType,
+)
+from awx_plugins.interfaces._temporary_private_django_api import (  # noqa: WPS436
+    gettext_noop,
+)
+
+from awx_plugins.credentials.injectors import (
```
webknjaz: Use relative imports internally, instead of relying on the top-level package name.
```suggestion
from ..injectors import (
```

src/awx_plugins/credentials/_managed_types/terraform.py
```diff
@@ -0,0 +1,43 @@
+"""HCP Terraform managed credential type."""
+# mypy: disable-error-code="assignment, misc"
```
webknjaz: The only reason some modules have global ignores is that at the time of the initial integration, MyPy was reporting many violations that couldn't be addressed immediately. This isn't a pattern to follow/extend for new modules.

Is this even necessary, though? If yes, add a FIXME comment explaining how to address these in the future.

But if at all possible, any new code shouldn't add new linting violations and have them addressed instead.

Finally, if we need to suppress violations, they should go inline, not be global for the entire module as this lets new problems sneak in and go unnoticed. Using `# type: ignore[<error-code>]` on the exact lines we need to suppress is much more granular. We should stick to using those as much as possible (if we can't actually fix things, that is!). Make sure to attach a code comment next to each suppression, explaining why it can't be fixed right now and what needs to happen for us to be able to get rid of the suppression.

src/awx_plugins/credentials/plugins.py
```diff
@@ -19,7 +20,7 @@
 )
 
 
-__all__ = ()  # noqa: WPS410
+__all__ = ('hcp_terraform',)  # noqa: WPS410
```
webknjaz: We shouldn't really make these public. The tuple is intentionally empty so that there's no public imports exposed.

src/awx_plugins/credentials/plugins.py
```diff
@@ -8,6 +8,7 @@
     gettext_noop,
 )
 
+from ._managed_types.terraform import hcp_terraform
```
webknjaz: Instead of re-exporting, point the entry points in the core packaging metadata (configured though `pyproject.toml` via PEP 621) to the new module directly. There's no reason to expose these objects here as well as this module isn't supposed to be importable directly, outside of the plugin loading mechanism.

src/awx_plugins/credentials/_managed_types/terraform.py
```diff
@@ -0,0 +1,45 @@
+"""HCP Terraform managed credential type."""
+
+from awx_plugins.interfaces._temporary_private_api import (  # noqa: WPS436
+    ManagedCredentialType,
+)
+from awx_plugins.interfaces._temporary_private_django_api import (  # noqa: WPS436
+    gettext_noop,
+)
+
+from ..injectors import hcp_terraform as hcp_terraform_injector  # noqa: TID252
```
webknjaz: @shvenkat-rh is this coming from our Ruff config? If yes, I'd rather disable the rule globally.

src/awx_plugins/credentials/_managed_types/terraform.py
```diff
@@ -0,0 +1,45 @@
...
+# ManagedCredentialType also returns Any
+hcp_terraform = ManagedCredentialType(  # type: ignore[misc]
+    namespace='hcp_terraform',
+    kind='cloud',
+    name=gettext_noop('HCP Terraform'),  # type: ignore[misc]
+    managed=True,
+    custom_injectors=hcp_terraform_injector,
+    inputs={
+        'fields': [  # type: ignore[misc]
+            {  # type: ignore[misc]
```
webknjaz, line 25: Are you sure these are needed?

src/awx_plugins/credentials/_managed_types/terraform.py
webknjaz: I suppose it's this one https://docs.astral.sh/ruff/rules/relative-imports/ ?
I've disabled it on `devel`. You can now drop the suppression and rebase.
```suggestion
from ..injectors import hcp_terraform as hcp_terraform_injector
```

src/awx_plugins/credentials/_managed_types/terraform.py
```diff
@@ -0,0 +1,45 @@
...
+                'id': 'hostname',
+                'label': gettext_noop('Hostname'),  # type: ignore[misc]
+                'type': 'string',
+                'help_text': gettext_noop(  # type: ignore[misc]
+                    'The hostname of your HCP Terraform instance (e.g., app.terraform.io)',
+                ),
+                'default': 'app.terraform.io',
+            },
+            {  # type: ignore[misc]
```
webknjaz, line 34: And this one?

webknjaz:
@shvenkat-rh if you post MyPy output, I can help decide if the suppressions are in the right place. Currently, it's not obvious if there's a justification for a few of those type ignores.

shvenkat-rh:
@webknjaz this is the MyPy output

```
$ git add -A && pre-commit run mypy --files src/awx_plugins/credentials/_managed_types/terraform.py 2>&1 | tail -80
                    'help_text': gettext_noop(
                                 ^
src/awx_plugins/credentials/_managed_types/terraform.py:33:13:41:13: error:
Expression type contains "Any" (has type "dict[str, Any]")  [misc]
                {
                ^
src/awx_plugins/credentials/_managed_types/terraform.py:35:17:35:50: error:
Expression type contains "Any" (has type "tuple[str, Any]")  [misc]
                    'label': gettext_noop('API Token'),
                    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
src/awx_plugins/credentials/_managed_types/terraform.py:35:26:35:50: error:
Expression has type "Any"  [misc]
                    'label': gettext_noop('API Token'),
                             ^~~~~~~~~~~~~~~~~~~~~~~~~
src/awx_plugins/credentials/_managed_types/terraform.py:38:17:40:17: error:
Expression type contains "Any" (has type "tuple[str, Any]")  [misc]
                    'help_text': gettext_noop(
                    ^
src/awx_plugins/credentials/_managed_types/terraform.py:38:30:40:17: error:
Expression has type "Any"  [misc]
                    'help_text': gettext_noop(
                                 ^
Generated Cobertura report: /Users/shvenkat/Downloads/awx-plugins/.tox/.tmp/.test-results/mypy--py-3.12/cobertura.xml
Generated HTML report (via XSLT): /Users/shvenkat/Downloads/awx-plugins/.tox/.tmp/.test-results/mypy--py-3.12/index.html
Generated TXT report (via XSLT): /Users/shvenkat/Downloads/awx-plugins/.tox/.tmp/.test-results/mypy--py-3.12/index.txt
Found 11 errors in 1 file (checked 20 source files)

MyPy, for Python 3.11....................................................Failed
- hook id: mypy
- exit code: 1

src/awx_plugins/credentials/_managed_types/terraform.py:23:19:42:9: error:
Expression type contains "Any" (has type "list[dict[str, Any]]")  [misc]
            'fields': [
                      ^
src/awx_plugins/credentials/_managed_types/terraform.py:24:13:32:13: error:
Expression type contains "Any" (has type "dict[str, Any]")  [misc]
                {
                ^
src/awx_plugins/credentials/_managed_types/terraform.py:26:17:26:49: error:
Expression type contains "Any" (has type "tuple[str, Any]")  [misc]
                    'label': gettext_noop('Hostname'),
                    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
src/awx_plugins/credentials/_managed_types/terraform.py:26:26:26:49: error:
Expression has type "Any"  [misc]
                    'label': gettext_noop('Hostname'),
                             ^~~~~~~~~~~~~~~~~~~~~~~~
src/awx_plugins/credentials/_managed_types/terraform.py:28:17:30:17: error:
Expression type contains "Any" (has type "tuple[str, Any]")  [misc]
                    'help_text': gettext_noop(
                    ^
src/awx_plugins/credentials/_managed_types/terraform.py:28:30:30:17: error:
Expression has type "Any"  [misc]
                    'help_text': gettext_noop(
                                 ^
src/awx_plugins/credentials/_managed_types/terraform.py:33:13:41:13: error:
Expression type contains "Any" (has type "dict[str, Any]")  [misc]
                {
                ^
src/awx_plugins/credentials/_managed_types/terraform.py:35:17:35:50: error:
Expression type contains "Any" (has type "tuple[str, Any]")  [misc]
                    'label': gettext_noop('API Token'),
                    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
src/awx_plugins/credentials/_managed_types/terraform.py:35:26:35:50: error:
Expression has type "Any"  [misc]
                    'label': gettext_noop('API Token'),
                             ^~~~~~~~~~~~~~~~~~~~~~~~~
src/awx_plugins/credentials/_managed_types/terraform.py:38:17:40:17: error:
Expression type contains "Any" (has type "tuple[str, Any]")  [misc]
                    'help_text': gettext_noop(
                    ^
src/awx_plugins/credentials/_managed_types/terraform.py:38:30:40:17: error:
Expression has type "Any"  [misc]
                    'help_text': gettext_noop(
                                 ^
```

webknjaz:
@shvenkat-rh okay, in that case I'd mention in the comment that nested `Any` structures cascade into making outer dicts/lists `Any` as well.

pyproject.toml
```diff
@@ -86,6 +86,7 @@ kubernetes_bearer_token = "awx_plugins.credentials.plugins:kubernetes_bearer_tok
 registry = "awx_plugins.credentials.plugins:registry"
 galaxy_api_token = "awx_plugins.credentials.plugins:galaxy_api_token"
 net = "awx_plugins.credentials.plugins:net"
+hcp_terraform = "awx_plugins.credentials._managed_types.terraform:hcp_terraform"
```
jessicamack: This only needs to be under the supported section.

src/awx_plugins/credentials/_managed_types/terraform.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+# FIXME: ManagedCredentialType and gettext_noop return Any from
+# _temporary_private_api interfaces. These nested Any structures cascade
+# into making outer dicts/lists typed as Any as well, requiring suppressions
+# throughout. Remove once those interfaces have proper type annotations.
+hcp_terraform = ManagedCredentialType(  # type: ignore[misc]
+    namespace='hcp_terraform',
+    kind='cloud',
+    name=gettext_noop('HCP Terraform'),  # type: ignore[misc]
```
jessicamack, line 20: Is there a reason for the gettext_noop uses? It's also causing the code coverage checks to fail.

shvenkat-rh: Agreed , removed gettext_noop and replaced with plain strings. This should fix the coverage issues.

jessicamack:
@shvenkat-rh lgtm have you already tried testing the plugin with a HCP credential?

src/awx_plugins/credentials/_managed_types/terraform.py
webknjaz: @jessicamack no, it does not cause the runtime coverage to fail. It only influences the typical coverage because it's uncovered. This is an internationalization/i18n helper and is a pre-existing pattern in the code base. This place probably shouldn't diverge from that.

shvenkat-rh: Thanks for the clarification. Restored gettext_noop to match the existing pattern.

pyproject.toml
shvenkat-rh: Line 89 [project.entry-points."awx_plugins.managed_credentials"], which AWX uses to discover and list credential types. Removing it hides hcp_terraform from AWX. I believe this should be retained

shvenkat-rh:
> @shvenkat-rh lgtm have you already tried testing the plugin with a HCP credential?

@jessicamack I have tested it without the pyproject.toml change you had suggest (When i try with the change you had suggested, awx fails to list the credential)

https://drive.google.com/file/d/1lemxGKg-QJF2o3Op5m1B7YvGJn6bdSUa/view?usp=sharing

webknjaz:
That's because awx can conditionally select which set of plugins to load automatically.

P.S. I've fixed up the failing job in `devel` so I recommend rebasing interactively and cleaning the commit list as it'd end up in the main tree post-merge. I've also split the runtime and typing coverage for better clarity — they are now being uploaded to two different services (runtime — Codecov; typing ­— Coveralls) and won't be overlayed anymore.

shvenkat-rh:
> That's because awx can conditionally select which set of plugins to load automatically.<

@webknjaz So do you suggest we retain `hcp_terraform = "awx_plugins.credentials._managed_types.terraform:hcp_terraform"` in [pyproject.toml](https://github.com/ansible/awx-plugins/pull/140/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711)  under `[project.entry-points."awx_plugins.managed_credentials"]` or not?

webknjaz:
No.

shvenkat-rh:
> No.

Okay, I am removing it. But while testing it out, I did retain it. Does that sound good to you?

webknjaz:
@shvenkat-rh the commit shouldn't claim that it made documentation changes that aren't there.

jessicamack (approved):

webknjaz (approved):

webknjaz:
> [!important]
> Merge Queue checks failed: https://github.com/ansible/awx-plugins/actions/runs/19468099427.

This is due to many services being down today as they are behind Cloudflare which returns HTTP 500 virtually everywhere: https://www.cloudflarestatus.com.

Scheduling the merge should be re-attempted once they recover. Closer to the evening, perhaps. Or tomorrow.

PR #147, opened 2026-01-20
Add Akeyless Secret Store/SSH Plugin

### LINKED ISSUE #148: Add Akeyless Secrets and SSH Plugin
https://github.com/ansible/awx/pull/16151

kgal-akl:
PR created as part of https://github.com/ansible/awx/pull/16151#issuecomment-3755856664.

## Summary by CodeRabbit

* **New Features**
  * Added Akeyless credential plugins (secret retrieval and SSH certificate issuance).

* **Tests**
  * Added comprehensive unit tests for secret formats, SSH issuance, TTL handling, and error paths.

* **Chores**
  * Added entry points and optional dependency groups to enable Akeyless integrations.
  * Relaxed type-checking for the Akeyless SDK to suppress import/type errors.

* **Style**
  * Updated linting and pre-commit tooling to include the new plugin.

* **Documentation**
  * Updated docs and spelling list to account for Akeyless names.


kgal-akl:
@jessicamack - PR is ready for review.

src/awx_plugins/credentials/akeyless.py
```diff
@@ -0,0 +1,430 @@
...
+    if not response.data:
+        raise RuntimeError(
+            'Failed to generate signed SSH certificate: no data returned.',
+        )
+    return response.data
+
+
+def akeyless_ssh_backend(**kwargs: Unpack[_AkeylessSshBackendKwargs]) -> str:
+    """Generate a signed SSH certificate using Akeyless."""
+    with CertFiles(kwargs.get('ca_cert') or None) as ca_cert_path:
```
webknjaz: `or None` seems redundant

src/awx_plugins/credentials/akeyless.py
```diff
@@ -0,0 +1,430 @@
...
+
+    def get_secret_value(self, req: object) -> Mapping[str, str]: ...
+
+    def get_ssh_certificate(
+        self,
+        req: object,
+    ) -> _SshCertResponse: ...
+
+
+common_plugin_inputs = [
```
webknjaz: could you make these global args private too?

src/awx_plugins/credentials/akeyless.py
```diff
@@ -0,0 +1,430 @@
+"""Akeyless credential plugins for AWX."""  # noqa: WPS202
+# pylint: disable=import-error, import-self, no-name-in-module
+# FIXME: the following violations must be addressed gradually and unignored
+# mypy: disable-error-code="import-not-found, import-untyped, no-untyped-def"
+
+import json
+from collections.abc import Mapping
+from typing import NotRequired, Protocol, TypedDict, Unpack, cast
```
webknjaz: Could you use these to make sure there's no leaking public names in the global namespace? (the typing ones are an emerging convention)
```suggestion
import collections.abc as _c
import json as _json
import typing as _t
```

src/awx_plugins/credentials/akeyless.py
```diff
@@ -0,0 +1,430 @@
+"""Akeyless credential plugins for AWX."""  # noqa: WPS202
+# pylint: disable=import-error, import-self, no-name-in-module
+# FIXME: the following violations must be addressed gradually and unignored
+# mypy: disable-error-code="import-not-found, import-untyped, no-untyped-def"
```
webknjaz: I like that this module mostly uses public names for global things but this seems to have way too many unjustified linting suppressions. I'd expect that any new modules would use inline suppressions for in justified cases, otherwise have the linting violations fixed instead. Similar global suppressions in other modules are there due to pre-existing issues at the migration time and fixing them wasn't a part of the migration (not fully, anyway).

Make sure that any inline suppressions have justification comments so the context persists where it's needed and can be discussed inline.

src/awx_plugins/credentials/akeyless.py
```diff
@@ -0,0 +1,430 @@
...
+        GetSSHCertificate(
+            token=token,
+            cert_issuer_name=ssh_inputs['cert_issue_name'],
+            cert_username=ssh_inputs['cert_username'],
+            ttl=_coerce_ttl(ssh_inputs.get('ttl')),
+            public_key_data=ssh_inputs['public_key_data'],
+        ),
+    )
+    if not response.data:
+        raise RuntimeError(
```
webknjaz: `RuntimeError` is good for interfacing with the plugin caller, in the top layer. But in the helper functions, it's probably a good idea to try using something more specific, like a `LookupError`, maybe. It can still be turned into a `RuntimeError` within the plugin functions, of course.

tests/importable_test.py
webknjaz, line 1: In addition to the smoke test, I'd expect some more functional tests to be bundled with new modules so there's regression coverage.

.flake8
```diff
@@ -118,6 +118,7 @@ per-file-ignores =
   tests/**.py: DAR, DCO020, S101, S105, S108, S404, S603, WPS202, WPS210, WPS430, WPS436, WPS441, WPS442, WPS450
 
   # The following ignores must be fixed and the entries removed from this config:
+  src/awx_plugins/credentials/akeyless.py: ANN101, WPS110, WPS202
```
webknjaz: What's the justification here? Can we avoid adding code with violations instead?

.mypy.ini
```diff
@@ -62,6 +62,14 @@ warn_unused_ignores = true
 # crashes with some decorators like `@functools.cache`:
 disallow_any_expr = false
 
+[mypy-awx_plugins.credentials.akeyless]
+# crashes with some decorators like `@functools.cache`:
+disallow_any_expr = false
+disallow_any_unimported = false
+disallow_any_explicit = false
+disallow_untyped_calls = false
+warn_return_any = false
```
webknjaz: Are all of these justified? It seems like the new module attempts to provide good type annotations so if there's any problems remaining, we should prefer surgical inline suppressions.

webknjaz:
The CI didn't start for some reason (or the runs got outdated and GH cleared them?). Re-opening the PR to trigger the CI.

webknjaz:
> [!tip]
> I wonder if this *needs to be* a part of `awx-plugins-core`. I don't see why it can't be its own project that end-users would opt-in to use via `pip-install`.

src/awx_plugins/credentials/akeyless.py
kgal-akl: You're right — `CertFiles.__enter__` already guards against falsy values with `if not self.cert: return None`, so `or None` is redundant. Fixed in both `akeyless_backend` and `akeyless_ssh_backend`.

src/awx_plugins/credentials/akeyless.py
kgal-akl: Good call. Renamed `common_plugin_inputs`, `akeyless_inputs`, and `akeyless_ssh_inputs` to `_common_plugin_inputs`, `_akeyless_inputs`, and `_akeyless_ssh_inputs`. Also made the three `frozenset` module-level constants private while I was at it, since they are implementation details.

src/awx_plugins/credentials/akeyless.py
kgal-akl: Done — changed to `import collections.abc as _c`, `import json as _json`, and `import typing as _t`, and updated all usages throughout the module.

src/awx_plugins/credentials/akeyless.py
kgal-akl: Fixed. The module-level `# mypy: disable-error-code=...` was redundant with the existing `[mypy-awx_plugins.credentials.akeyless]` section in `.mypy.ini` — removed entirely. The `# pylint: disable=import-error,no-name-in-module` is now a targeted block around only the akeyless SDK imports, with a comment explaining the reason (the SDK has no type stubs). The erroneous `import-self` suppression was also dropped. The `# noqa: WPS202` stays on the module docstring — with an explanatory comment below it — since the two-plugin structure genuinely exceeds the member threshold by design.

src/awx_plugins/credentials/akeyless.py
kgal-akl: Good suggestion. `_authenticate` and `_fetch_ssh_certificate` now raise `ValueError` internally. The top-level `akeyless_backend` and `akeyless_ssh_backend` each catch both `ApiException` and `ValueError` in a single `try` block and wrap them into `RuntimeError` for the plugin caller. I also moved the `_authenticate` call inside the `try` block so auth failures are covered by the same wrapping.

tests/importable_test.py
kgal-akl: Added `tests/akeyless_test.py` with 13 tests covering: text/generic, JSON with key, and password sub-type secrets; unsupported item types; `ApiException` wrapping to `RuntimeError`; auth failures; missing JSON keys; SSH certificate success and failure paths (no data returned); API exceptions on SSH calls; and `_coerce_ttl` for all valid cases (`None`, empty string, int, string-int) and an invalid input.

.flake8
kgal-akl: Fixed. `ANN101` is no longer needed — Protocol methods in `_AkeylessApi` now annotate `self: _t.Self` (valid in Python 3.11). `WPS110` is now an inline `# noqa: WPS110` on the `data` attribute of `_SshCertResponse` with a comment that it must match the SDK response attribute name. `WPS202` is an inline `# noqa: WPS202` on the module docstring with an explanatory comment. The `akeyless.py` entry has been removed from `.flake8`'s per-file-ignores entirely.

.mypy.ini
kgal-akl: Updated. The copy-pasted `@functools.cache` comment has been replaced with an accurate explanation: all suppressions in this section exist because the akeyless SDK does not ship type stubs, so mypy cannot resolve its types. They will remain until the SDK gains proper typing support.

kgal-akl:
Re the tip about whether this needs to be part of `awx-plugins-core`: we'd prefer to keep it here to stay consistent with how other third-party integrations (Azure Key Vault, HashiVault, CyberArk AIM, Conjur, etc.) are packaged. Being in the core repo gives it the same CI, linting, and test infrastructure — and makes it discoverable to AWX users without an additional install step. Happy to discuss further if there's a strong reason to break it out.

webknjaz:
Well, the maintenance resources (human-wise) are quite limited here. And so the advantages can easily end up being downsides. As for the infra — it can be copied to whatever place necessary :man_shrugging:

src/awx_plugins/credentials/akeyless.py
```diff
@@ -0,0 +1,441 @@
+"""Akeyless credential plugins for AWX."""  # noqa: WPS202
+# WPS202 is expected: implementing two credential plugins requires Protocol
```
webknjaz: Does this suppression actually work, though? IIRC, the WPS plugin would normally not allow in-file suppressions for such global checks (they might've fixed that in newer versions but I haven't checked)

src/awx_plugins/credentials/akeyless.py
```diff
@@ -0,0 +1,441 @@
...
+
+
+class _AkeylessApi(_t.Protocol):
+    def auth(self: _t.Self, auth: object) -> _AuthResponse: ...
+
+    def describe_item(self: _t.Self, req: object) -> _DescribeItemResponse: ...
+
+    def get_secret_value(
+        self: _t.Self,
+        req: object,
```
webknjaz: Why can't this be `GetSecretValue`? I see it's what's passed in runtime.

src/awx_plugins/credentials/akeyless.py
```diff
@@ -0,0 +1,441 @@
...
+    Configuration,
+    DescribeItem,
+    GetSecretValue,
+    V2Api,
+)
+from akeyless.models.get_ssh_certificate import GetSSHCertificate
+from akeyless.rest import ApiException
+# pylint: enable=import-error,no-name-in-module
+
+from .plugin import CertFiles, CredentialPlugin
```
webknjaz: Can we
```suggestion
from . import plugin as _plugin
```
?

src/awx_plugins/credentials/akeyless.py
```diff
@@ -0,0 +1,441 @@
...
+from akeyless import (
+    ApiClient,
+    Auth,
+    Configuration,
+    DescribeItem,
+    GetSecretValue,
+    V2Api,
+)
+from akeyless.models.get_ssh_certificate import GetSSHCertificate
+from akeyless.rest import ApiException
```
webknjaz: With this import style, all these things end up in the global namespace. It's probably a good idea to make them private or declare an explicit `__all__`, perhaps.

tests/akeyless_test.py
```diff
@@ -0,0 +1,296 @@
...
+
+import pytest
+from pytest_mock import MockerFixture
+
+from akeyless.rest import ApiException  # pylint: disable=import-error
+
+from awx_plugins.credentials import akeyless as akeyless_mod
+
+
+_SECRET_PATH = '/test/secret'
```
webknjaz, line 15: FTR, I'm not as worried about the private names in the test modules since they aren't public API by definition. It's the fact that the runtime modules can end up something that somebody would assume public that I care about.

tests/akeyless_test.py
```diff
@@ -0,0 +1,296 @@
...
+    token: str | None = 'test-token',
+    item_type: str = 'STATIC_SECRET',
+    item_sub_type: str = 'generic',
+    secret_format: str = 'text',
+    secret_path: str = _SECRET_PATH,
+    secret_data: str = 'my-secret-value',
+    ssh_cert_data: str | None = 'ssh-rsa-SIGNED-CERT',
+) -> object:
+    """Create a mock Akeyless API instance with configurable responses."""
+    mock_static_info = mocker.MagicMock()
```
webknjaz: Does it have to be magic or would a regular mock be enough?

tests/akeyless_test.py
```diff
@@ -0,0 +1,296 @@
...
+    mocker: MockerFixture,
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    """Retrieve a specific key from a JSON-format structured secret."""
+    mock_api = _make_mock_api(
+        mocker,
+        secret_format='json',
+        secret_data='{"db_password": "s3cr3t", "db_user": "admin"}',
+    )
+    monkeypatch.setattr(akeyless_mod, '_setup_client', lambda *_: mock_api)
```
webknjaz: This could probably be wrapped into a pytest fixture with args passed through parametrize+indirect.

tests/akeyless_test.py
```diff
@@ -0,0 +1,296 @@
...
+
+    result = akeyless_mod.akeyless_backend(  # type: ignore[arg-type]
+        **_backend_kwargs(secret_key='db_password'),
+    )
+
+    assert result == 's3cr3t'
+
+
+def test_akeyless_backend_password_secret_username_key(
```
webknjaz: It appears that this test is the same as the previous one structurally? In such cases, having one test function is enough — you just need to use `@pytest.mark.parametrize` to make it into distinct test instances within pytest (and in the test report/output).

tests/akeyless_test.py
```diff
@@ -0,0 +1,296 @@
...
+        pytest.param(3600, 3600, id='int'),
+        pytest.param('7200', 7200, id='string-int'),
+    ),
+)
+def test_coerce_ttl_valid_inputs(
+    ttl_input: int | str | None,
+    expected: int | None,
+) -> None:
+    """_coerce_ttl should convert numeric-like values and pass None/'' through."""
+    assert akeyless_mod._coerce_ttl(ttl_input) == expected  # noqa: WPS437
```
webknjaz, line 344: Could you leave a comment mentioning what the linting rule number means?

src/awx_plugins/credentials/akeyless.py
```diff
@@ -0,0 +1,441 @@
...
+            token = _authenticate(
+                api_instance,
+                kwargs['access_id'],
+                kwargs['access_key'],
+            )
+            return _fetch_ssh_certificate(
+                api_instance,
+                token,
+                kwargs,
+            )
```
webknjaz, line 447: It's usually best to have a single instruction in try-blocks so that it's clear where the exception is expected, ans which one. If you want to wrap the entire thing, perhaps, we could move it into an external abstraction layer via something like `@contextmanager.contextlib` — it'd also make this reusable as I see that it duplicates the other plugin.

In fact, I was just making a suggestion in another PR (https://github.com/ansible/awx-plugins/pull/176#discussion_r2952528042) that maybe the token retrieval could leave in a decorator that wraps the function with a context manager and passes it as an argument. Check it out — maybe, it'll help your case too: https://github.com/ansible/awx-plugins/pull/176/commits/441c4d9be407455607f76957fc996f85faa4da85#diff-6fc6d0b42d0d2b46c77c198dc408d6d225dccbce124d813e5c3fc22cab84a8f3R533-R534. Though, in this PR, I'd also move the `with`-block in there.

src/awx_plugins/credentials/akeyless.py
```diff
@@ -0,0 +1,441 @@
...
+        except ApiException as exc:
+            raise RuntimeError(
+                f'Akeyless API error: {exc.reason} (Status: {exc.status})',
+            ) from exc
+        except ValueError as exc:
```
webknjaz: *(personal preference)* I like naming things explicitly, including the exception intances. So here's I'd go for `val_err` and `api_exc`, or something like that, if you don't mind.

tests/akeyless_test.py
```diff
@@ -0,0 +1,296 @@
...
+    mock_api = _make_mock_api(mocker, ssh_cert_data='signed-cert')
+    monkeypatch.setattr(akeyless_mod, '_setup_client', lambda *_: mock_api)
+
+    result = akeyless_mod.akeyless_ssh_backend(  # type: ignore[arg-type]
+        **_ssh_kwargs(ttl=3600),
+    )
+
+    assert result == 'signed-cert'
+    _, call_kwargs = mock_api.get_ssh_certificate.call_args
+    assert call_kwargs['body'].ttl == 3600
```
webknjaz: https://github.com/ansible/awx-plugins/actions/runs/23249666391/job/67597596597?pr=147#step:20:100

src/awx_plugins/credentials/akeyless.py
```diff
@@ -0,0 +1,441 @@
...
+
+import collections.abc as _c
+import json as _json
+import typing as _t
+
+from awx_plugins.interfaces._temporary_private_django_api import (  # noqa: WPS436
+    gettext_noop as _,
+)
+
+# The akeyless SDK does not ship type stubs; pylint cannot resolve its imports.
```
webknjaz: `pylint`'s got a point (kinda) — this may be a recursive self-import in some contexts: https://github.com/ansible/awx-plugins/actions/runs/23249666391/job/67597596573?pr=147#step:16:996. It's usually a good idea to use a different module name than a top-level name of some other namespace. OTOH, maybe https://github.com/pylint-dev/pylint/issues/10147 will fix it but I'm not sure about the timeline of that happening. This may end up being a blocker FTR.

webknjaz: Oh, wait. You didn't update `.pre-commit-config.yaml` — try that and it might make the problem go away. In the end, you shouldn't need to suppress the errors here.
Make sure to update both the pylint env and the MyPy envs.

src/awx_plugins/credentials/akeyless.py
webknjaz: Fixing this will also address the typing errors in the test module.

src/awx_plugins/credentials/akeyless.py
```diff
@@ -0,0 +1,441 @@
+"""Akeyless credential plugins for AWX."""  # noqa: WPS202
```
webknjaz: To fix https://github.com/ansible/awx-plugins/actions/runs/23249666391/job/67597596601?pr=147#step:16:97, add an entry somewhere around here: https://github.com/ansible/awx-plugins/blob/c464cf276b9344056d7f47090bf361c2a8d85b45/docs/spelling_wordlist.txt#L1.

src/awx_plugins/credentials/akeyless.py
```diff
@@ -0,0 +1,441 @@
...
+import collections.abc as _c
+import json as _json
+import typing as _t
+
+from awx_plugins.interfaces._temporary_private_django_api import (  # noqa: WPS436
+    gettext_noop as _,
+)
+
+# The akeyless SDK does not ship type stubs; pylint cannot resolve its imports.
+# pylint: disable=import-error,no-name-in-module
```
webknjaz: If nothing else helps (adding the dep into the linting envs should!), I'd consider checking if https://pylint.pycqa.org/en/stable/user_guide/configuration/all-options.html#ignored-modules would do.

webknjaz:
@kgal-akl I think I've posted all the feedback I could. But the ultimate decision on this belonging here would be by the respective team. I'm just helping out with the repo infra occasionally.

src/awx_plugins/credentials/akeyless.py
kgal-akl: https://github.com/ansible/awx-plugins/pull/147/commits/2ee90140faad7a7017f042be2e79623ee4317c0a

src/awx_plugins/credentials/akeyless.py
kgal-akl: https://github.com/ansible/awx-plugins/pull/147/commits/2ee90140faad7a7017f042be2e79623ee4317c0a

tests/akeyless_test.py
kgal-akl: https://github.com/ansible/awx-plugins/pull/147/commits/2ee90140faad7a7017f042be2e79623ee4317c0a

src/awx_plugins/credentials/akeyless.py
kgal-akl: https://github.com/ansible/awx-plugins/pull/147/commits/2ee90140faad7a7017f042be2e79623ee4317c0a

src/awx_plugins/credentials/akeyless.py
kgal-akl: https://github.com/ansible/awx-plugins/pull/147/commits/2ee90140faad7a7017f042be2e79623ee4317c0a

tests/akeyless_test.py
kgal-akl: https://github.com/ansible/awx-plugins/pull/147/commits/2ee90140faad7a7017f042be2e79623ee4317c0a

tests/akeyless_test.py
kgal-akl: https://github.com/ansible/awx-plugins/pull/147/commits/2ee90140faad7a7017f042be2e79623ee4317c0a

tests/akeyless_test.py
kgal-akl: https://github.com/ansible/awx-plugins/pull/147/commits/2ee90140faad7a7017f042be2e79623ee4317c0a

tests/akeyless_test.py
kgal-akl: https://github.com/ansible/awx-plugins/pull/147/commits/2ee90140faad7a7017f042be2e79623ee4317c0a

tests/akeyless_test.py
kgal-akl: https://github.com/ansible/awx-plugins/pull/147/commits/2ee90140faad7a7017f042be2e79623ee4317c0a

src/awx_plugins/credentials/akeyless.py
kgal-akl: https://github.com/ansible/awx-plugins/pull/147/commits/2ee90140faad7a7017f042be2e79623ee4317c0a

src/awx_plugins/credentials/akeyless.py
kgal-akl: https://github.com/ansible/awx-plugins/pull/147/commits/2ee90140faad7a7017f042be2e79623ee4317c0a

src/awx_plugins/credentials/akeyless.py
kgal-akl: https://github.com/ansible/awx-plugins/pull/147/commits/2ee90140faad7a7017f042be2e79623ee4317c0a

src/awx_plugins/credentials/akeyless.py
kgal-akl: https://github.com/ansible/awx-plugins/pull/147/commits/2ee90140faad7a7017f042be2e79623ee4317c0a

.mypy.ini
```diff
@@ -106,8 +115,23 @@ disallow_any_expr = false
...
 [mypy-tests.*]
 # crashes with some decorators like `@pytest.mark.parametrize`:
 disallow_any_expr = false
 # fails on `@hypothesis.given()`:
 disallow_any_decorated = false
+# fixture return types like `Callable[..., object]` use `...` as Any:
+disallow_any_explicit = false
+
+[mypy-tests.akeyless_test]
+# Mock API objects are typed as `object`; test kwargs use generic dict
```
webknjaz: I'd rather have these ignores inline

.pylintrc.toml
```diff
@@ -106,7 +106,11 @@ py-version = "3.11"
 # source root is an absolute path or a path relative to the current working
 # directory used to determine a package namespace for modules located under the
 # source root.
-# source-roots =
+# Setting src/ prevents pylint from adding individual credential module
+# directories to sys.path, which would cause false "module imports itself"
+# warnings and circular-import errors for files with the same name as their
+# third-party dependencies (e.g. akeyless.py vs the akeyless SDK):
+source-roots = ["src"]
```
webknjaz, line 112: This is already achieved by the `init-hook` on line 64. I've been meaning to migrate to `source-roots` but didn't get to it. It's literally among the experiments in my Git checkout right now:
```diff
diff --git a/.pylintrc.toml b/.pylintrc.toml
index 3ffdc0e837..39b177a8d8 100644
--- a/.pylintrc.toml
+++ b/.pylintrc.toml
@@ -61,18 +61,19 @@ ignore-patterns = ["^\\.#"]
 # This patch injects the project directory into the import path so that the
 # local `pytest` plugin can be imported when `pylint-pytest` invokes it when
 # exploring the fixtures available:
-init-hook = """
-import os, pathlib, sys
-repo_root_path = pathlib.Path.cwd()
-src_path = repo_root_path / 'src'
-sys.path[:0] = [str(src_path if src_path.exists() else repo_root_path)]
-os.environ['PYTHONPATH'] = sys.path[0]
-"""
+# init-hook = """
+# import os, pathlib, sys
+# repo_root_path = pathlib.Path.cwd()
+# src_path = repo_root_path / 'src'
+# sys.path[:0] = [str(src_path if src_path.exists() else repo_root_path)]
+# os.environ['PYTHONPATH'] = sys.path[0]
+# """
 
 # Use multiple processes to speed up Pylint. Specifying 0 will auto-detect the
 # number of processors available to use, and will cap the count on Windows to
 # avoid hangs.
-jobs = 0
+# jobs = 0
+jobs = 1
 
 # Control the amount of potential inferred values when inferring a single object.
 # This can help the performance when dealing with large functions or complex,
@@ -107,6 +108,9 @@ py-version = "3.11"
 # directory used to determine a package namespace for modules located under the
 # source root.
 # source-roots =
+source-roots = [
+  "src/",
+]
 
 # Allow loading of arbitrary C extensions. Extensions are imported into the
 # active Python interpreter and may run arbitrary code.
@@ -435,7 +439,7 @@ disable = [
   "disallowed-name",
   "duplicate-code",
   "fixme",
-  "import-outside-toplevel",
+  # "import-outside-toplevel",
   "invalid-name",
   "line-too-long",
   "missing-class-docstring",
@@ -444,11 +448,11 @@ disable = [
   "missing-timeout",
   "no-else-return",
   "no-member",
-  "no-name-in-module",  # false-positive: https://github.com/pylint-dev/pylint/issues/10147#issuecomment-3946199493
+  # "no-name-in-module",  # false-positive: https://github.com/pylint-dev/pylint/issues/10147#issuecomment-3946199493
   "no-self-use",
   "pointless-string-statement",
   "raise-missing-from",
-  "relative-beyond-top-level",
+  # "relative-beyond-top-level",
   "singleton-comparison",
   "too-few-public-methods",
   "too-many-branches",
```


But regardless, this seems like something that doesn't belong in this PR but could be submitted separately as a linting config / infra improvement rather than related to a plugin being presented. Would you like to send another PR with just this migration? (the `init-hook`-to-`source-roots` parts of the snippet)

.mypy.ini
```diff
@@ -106,8 +115,23 @@ disallow_any_expr = false
...
 disallow_any_expr = false
 
+[mypy-akeyless]
+# The akeyless SDK does not ship type stubs:
+ignore_missing_imports = true
+
+[mypy-akeyless.*]
+# The akeyless SDK does not ship type stubs:
+ignore_missing_imports = true
+
```
webknjaz: The preferred solution is adding own stubs into the Git tree like https://github.com/ansible/awx_plugins.interfaces/tree/1cec8f60cbb84806ce5cb08c27c301dd86dc866a/_type_stubs.

MyPy's already configured to pick them up from the same dir: https://github.com/ansible/awx-plugins/blob/c464cf276b9344056d7f47090bf361c2a8d85b45/.mypy.ini#L36C69-L36C104.

.mypy.ini
```diff
@@ -62,6 +62,15 @@ warn_unused_ignores = true
 # crashes with some decorators like `@functools.cache`:
 disallow_any_expr = false
 
+[mypy-awx_plugins.credentials.akeyless]
+# The akeyless SDK does not ship type stubs. These suppressions are needed
```
webknjaz: All of them? Why?

.flake8
```diff
@@ -117,6 +117,11 @@ per-file-ignores =
   # additionally test docstrings don't need param lists (DAR, DCO020):
   tests/**.py: DAR, DCO020, S101, S105, S108, S404, S603, WPS202, WPS210, WPS430, WPS436, WPS441, WPS442, WPS450
 
+  # The following ignores have been researched and should be considered permanent:
+  # WPS202: two-plugin module (secret store + SSH) with Protocol types, TypedDicts,
+  # and multiple helpers per plugin; restructuring would harm readability.
```
webknjaz: Not necessarily. There's no reason why a single module can't be turned into a folder with some separation of common helpers and entry points, and maybe types.

pytest.ini
```diff
@@ -1,4 +1,8 @@
 [pytest]
+# Add src/ to sys.path so tests can import the package even in environments
+# where `pip install -e .` has not been run (e.g. the pre-commit pylint venv):
+pythonpath = src
```
webknjaz: This is definitely a wrong infra change. Things aren't supposed to be imported from the tree, only from installed envs. Additionally, `pytest` isn't supposed to be run directly.

.mypy.ini
```diff
@@ -106,8 +115,23 @@ disallow_any_expr = false
...
 # crashes with some decorators like `@pytest.mark.parametrize`:
 disallow_any_expr = false
 # fails on `@hypothesis.given()`:
 disallow_any_decorated = false
+# fixture return types like `Callable[..., object]` use `...` as Any:
+disallow_any_explicit = false
+
+[mypy-tests.akeyless_test]
+# Mock API objects are typed as `object`; test kwargs use generic dict
+# rather than the specific TypedDicts the plugin functions expect:
```
webknjaz: Can't they be casted to whatever type necessary? You've done this in the runtime module already. No need to have multiple approaches for the same thing..

src/awx_plugins/credentials/akeyless.py
```diff
@@ -0,0 +1,464 @@
...
+import json as _json
+import typing as _t
+from collections.abc import Mapping as _Mapping
+
+from awx_plugins.interfaces._temporary_private_django_api import (  # noqa: WPS436
+    gettext_noop as _,
+)
+
+# The akeyless SDK does not ship type stubs; pylint cannot resolve its imports.
+# When not installed, pylint misidentifies `from akeyless import` as a
```
webknjaz: So it should be installed, just like all the other deps, not need to have an exception for this one.

tests/akeyless_test.py
webknjaz: That looks unrelated.

tests/akeyless_test.py
webknjaz: I don't see how this is connected.

webknjaz: Also, don't forget about the specs.

tests/akeyless_test.py
```diff
@@ -0,0 +1,304 @@
...
+    mock_describe.item_type = item_type
+    mock_describe.item_sub_type = item_sub_type
+    mock_describe.item_general_info = mock_general_info
+
+    mock_api = mocker.MagicMock()
+    mock_api.auth.return_value.token = token
+    mock_api.describe_item.return_value = mock_describe
+    mock_api.get_secret_value.return_value = {secret_path: secret_data}
+    mock_api.get_ssh_certificate.return_value.data = ssh_cert_data
+    return mock_api
```
webknjaz, line 85: By the way, wouldn't it be easier to go with stubs instead of mocks? It seems like half of these would be just fine with a dumb dataclass-based implementation instead of mocks (let alone magic ones).

webknjaz:
@kgal-akl plz avoid adding merge commits in favor of rebasing.

PR #150, opened 2026-02-11
Add a smoke test for plugin input IDs in hashivault credentials

PabloHiro:
AAP-64507

- Moved hashivault tests from `tests/credential_plugins_test.py` to a new file `tests/hashivault_test.py`.
- Added tests that validate the existing fields and metadata of the credentials in `tests/hashivault_test.py` in preparation for upcoming PRs

## Summary by CodeRabbit

* **Tests**
  * Added a comprehensive test suite for HashiCorp Vault credential handling, covering multiple auth methods and input validation.
  * Removed redundant HashiCorp-specific tests from the consolidated credential plugin tests to refocus that file on the aim module.
  * Retained aim-related tests and legacy import checks for Azure-related imports.


TESTING.md
```diff
@@ -0,0 +1,40 @@
...
+# Install tox
+pip install tox
+
+# Run the full test suite
+tox r -e py
+
+# Run tests for a specific Python version
+tox r -e py311
+tox r -e py312
+tox r -e py313
```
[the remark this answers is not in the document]
PabloHiro: Python 3.11 is needed for `pre-commit` to work properly. Tests are compatible with different Python versions. Not an issue

tests/hashivault_test.py
```diff
@@ -0,0 +1,240 @@
+"""Tests for HashiCorp Vault credential plugins."""
+
+from typing import Any, Self
```
webknjaz: The latest best practice is to do
```suggestion
import typing as _t
```
I'll add a linter for this later. Also, avoid `_t.Any` at all cost. Whenever possible, MyPy's docs recommend using `object` instead.

tests/hashivault_test.py
```diff
@@ -0,0 +1,240 @@
...
+from unittest import mock
```
webknjaz: Once you rebase, this will be blocked as #155 now correctly forbids such imports.

tests/hashivault_test.py
```diff
@@ -0,0 +1,240 @@
...
+
+import pytest
+
+from awx_plugins.credentials import hashivault
+
+
+class TestHashivaultAuthMethods:
```
webknjaz: I recommend not overindenting tests with classes since you don't use the shared state. It's common when using stdlib's `unittest` or Java but in `pytest`, you can just put functions into separate modules for namespacing/grouping. You may want to start adding sub-folders under `tests/`, perhaps mirror some of the layout under `src/`.

tests/hashivault_test.py
```diff
@@ -0,0 +1,240 @@
...
+    def test_client_cert_auth_no_role(self: Self) -> None:
+        """Test client_cert_auth without role returns None name."""
+        kwargs: dict[str, str] = {}
+        expected_res = {
+            'name': None,
+        }
+        res = hashivault.client_cert_auth(  # type: ignore[no-untyped-call]
+            **kwargs,
+        )
+        assert res == expected_res
```
webknjaz: These are the same test. Use `@pytest.mark.parametrize`.

tests/hashivault_test.py
webknjaz: Though, getting rid of classes will likely make this redundant.

tests/hashivault_test.py
```diff
@@ -0,0 +1,240 @@
...
+            token = hashivault.handle_auth(  # type: ignore[no-untyped-call]
+                **kwargs,
+            )
+            method_mock.assert_called_with(**kwargs, auth_param=auth_params)
+            assert token == 'the_token'
+
+    def test_handle_auth_not_enough_args(self: Self) -> None:
+        """Test handle_auth raises exception without valid auth method."""
+        with pytest.raises(
+            Exception,
```
webknjaz: Checking for catch-all exceptions isn't very useful. Best be specific.

tests/hashivault_test.py
```diff
@@ -0,0 +1,240 @@
...
+            match=r'authentication parameters must be set',
```
webknjaz: Best use `^` / `$` markers to make it clear that this is a regex. Otherwise, this tends to confuse people.

tests/hashivault_test.py
```diff
@@ -0,0 +1,240 @@
...
+        ):
+            hashivault.handle_auth()  # type: ignore[no-untyped-call]
+
+
+class TestHashivaultKvPlugin:
+    """Tests for the HashiCorp Vault KV credential plugin."""
+
+    def test_fields(self: Self) -> None:
+        """Verify input fields are present with expected IDs."""
+        inputs: dict[  # type: ignore[explicit-any]
```
webknjaz: It'd be good to define a typed dict for this.

tests/hashivault_test.py
```diff
@@ -0,0 +1,240 @@
...
+        ] = hashivault.hashivault_kv_plugin.inputs  # type: ignore[assignment]
+        metadata_ids = [metadata['id'] for metadata in inputs['metadata']]
+        expected_metadata = [
+            'secret_backend',
+            'secret_path',
+            'auth_path',
+            'secret_key',
+            'secret_version',
+        ]
+        assert metadata_ids == expected_metadata
```
webknjaz: These are pretty much the same test. Use `parametrize` instead of shape duplication.

tests/hashivault_test.py
```diff
@@ -0,0 +1,240 @@
...
+        ] = hashivault.hashivault_ssh_plugin.inputs  # type: ignore[assignment]
+        metadata_ids = [metadata['id'] for metadata in inputs['metadata']]
+        expected_metadata = [
+            'public_key',
+            'secret_path',
+            'auth_path',
+            'role',
+            'valid_principals',
+        ]
+        assert metadata_ids == expected_metadata
```
webknjaz: Same as the above.

webknjaz (changes requested):

tests/hashivault_test.py
PabloHiro: The source code uses `Exception` straight away

tests/hashivault_test.py
PabloHiro: You got rid of this one in `devel` :/

webknjaz:
@PabloHiro check out https://gist.github.com/webknjaz/cb7d7bf62c3dda4b1342d639d0e78d79 to learn how to write good commit messages. TL;DR titles are imperative unprefixed summaries for atomic changes and the descriptions explain motivation / considerations etc. explain the "why" / unobvious side effects. I recommend running `gitlint` for self-checks.

tests/hashivault_test.py
webknjaz: Ouch.. That's terrible!

tests/unit/credentials/hashivault_test.py
webknjaz, line 1: Let's move this to `tests/credentials/hashivault_test.py`, mirroring the main package.

tests/unit/credentials/hashivault_test.py
```diff
@@ -0,0 +1,229 @@
...
+    kwargs = {
+        'role_id': 'the_role_id',
+        'secret_id': 'the_secret_id',
+    }
+    expected_res = {
+        'role_id': 'the_role_id',
+        'secret_id': 'the_secret_id',
+    }
+    res = hashivault.approle_auth(**kwargs)  # type: ignore[no-untyped-call]
+    assert res == expected_res
```
webknjaz, line 21: You don't really need to copies of the same data.
```suggestion
    auth_kwargs = {
        'role_id': 'the_role_id',
        'secret_id': 'the_secret_id',
    }
    resolved_kwargs = hashivault.approle_auth(**auth_kwargs)  # type: ignore[no-untyped-call]
    assert resolved_kwargs == auth_kwargs
```

I wish there could be integration testing included rather than a simple thing like this..

tests/unit/credentials/hashivault_test.py
```diff
@@ -0,0 +1,229 @@
...
+def test_hashivault_kubernetes_auth(mocker: MockerFixture) -> None:
+    """Test kubernetes_auth."""
+    kwargs = {
+        'kubernetes_role': 'the_kubernetes_role',
+    }
+    expected_res = {
+        'role': 'the_kubernetes_role',
+        'jwt': 'the_jwt',
+    }
+    path_mock = mocker.patch('pathlib.Path')
```
webknjaz, line 33: Always use `autospec=True` when patching stuff.
```suggestion
    path_mock = mocker.patch('pathlib.Path', autospec=True)
```

tests/unit/credentials/hashivault_test.py
```diff
@@ -0,0 +1,229 @@
...
+    res = hashivault.kubernetes_auth(  # type: ignore[no-untyped-call]
+        **kwargs,
+    )
+    path_mock.assert_called_with(
+        '/var/run/secrets/kubernetes.io/serviceaccount/token',
+    )
+    assert res == expected_res
+
+
+def test_hashivault_client_cert_auth_explicit_role() -> None:  # noqa: WPS118
```
webknjaz, line 44: Any `noqa` must be accompanied with a code comment with justification. The readers don't normally have all the codes from all the plugins memorized to understand this w/o unnecessary burden.

tests/unit/credentials/hashivault_test.py
```diff
@@ -0,0 +1,229 @@
...
+    }
+    path_mock = mocker.patch('pathlib.Path')
+    path_mock.return_value.open = mocker.mock_open(read_data='the_jwt')
+    res = hashivault.kubernetes_auth(  # type: ignore[no-untyped-call]
+        **kwargs,
+    )
+    path_mock.assert_called_with(
+        '/var/run/secrets/kubernetes.io/serviceaccount/token',
+    )
+    assert res == expected_res
```
webknjaz, line 41: This is *way too much* mocking: https://news.ycombinator.com/item?id=45949102 / https://nedbatchelder.com/blog/202511/why_your_mock_breaks_later.

Best make the module testable so there's fewer fragile hacks.

Additionally, have the expectations/motivation in docstrings: not just "I added a test so a test has been added and thus you see a test".

This is solvable by moving the hardcoded path in `hashivault.py` to a module-global `_k8s_jwt_path` variable, and using it there. Then here, we can substitute it with a temporarily created path, write our sentinel there w/o manipulating the stdlib: don't mock what you don't own — https://hynek.me/articles/what-to-mock-in-5-mins/.

```suggestion
def test_hashivault_kubernetes_auth(monkeypatch: pytest.MonkeyPatch, tmp_path: pathlib.Path) -> None:
    """Test ``kubernetes_auth()`` returns a stripped token from file."""
    jwt_token_file_path = tmp_path / 'plain-text-jwt'
    pretend_jwt_token_value = 'the_jwt'
    pretend_kubernetes_role = 'the_kubernetes_role'
    jwt_token_file_path.write_text(f'{pretend_jwt_token_value}    ')  # rstrip
    monkeypatch.setattr(hashivault, '_k8s_jwt_path', jwt_token_file_path)
 
    expected_auth_data = {
        'role': pretend_kubernetes_role,
        'jwt': pretend_jwt_token_value,
    }
    resolved_auth_data = hashivault.kubernetes_auth(  # type: ignore[no-untyped-call]
        kubernetes_role=pretend_kubernetes_role,
    )
    assert resolved_auth_data == expected_auth_data
```

tests/unit/credentials/hashivault_test.py
```diff
@@ -0,0 +1,229 @@
...
+    method_mock = mocker.patch.object(hashivault, 'method_auth')
+    method_mock.return_value = 'the_token'
+    token = hashivault.handle_auth(  # type: ignore[no-untyped-call]
+        **kwargs,
+    )
+    method_mock.assert_called_with(**kwargs, auth_param=kwargs)
+    assert token == 'the_token'
+
+
+def test_hashivault_handle_auth_kubernetes(mocker: MockerFixture) -> None:
```
webknjaz, line 102: This needs a similar treatment to `test_hashivault_kubernetes_auth`.

tests/unit/credentials/hashivault_test.py
```diff
@@ -0,0 +1,229 @@
...
+    }
+    res = hashivault.client_cert_auth(  # type: ignore[no-untyped-call]
+        **kwargs,
+    )
+    assert res == expected_res
+
+
+def test_hashivault_client_cert_auth_no_role() -> None:
+    """Test client_cert_auth with no role."""
+    kwargs: dict[str, str] = {}
```
webknjaz, line 60: You don't actually need this.

tests/unit/credentials/hashivault_test.py
```diff
@@ -0,0 +1,229 @@
...
+    expected_res = {
+        'name': None,
+    }
+    res = hashivault.client_cert_auth(  # type: ignore[no-untyped-call]
+        **kwargs,
+    )
```
webknjaz, line 66: This could be
```suggestion
    res = hashivault.client_cert_auth()  # type: ignore[no-untyped-call]
```
However it seems like there's a bunch of tests are the same test duplicated with different params. Like this one and `test_hashivault_client_cert_auth_explicit_role`. Collapse them into a single test parametrized.

tests/hashivault_test.py
```diff
@@ -0,0 +1,229 @@
...
+                'auth_path',
+                'role',
+                'valid_principals',
+            ],
+            id='ssh-metadata',
+        ),
+    ),
+)
+def test_plugin_input_ids(
+    plugin: object,
```
webknjaz: AFAICS, it's actually
```suggestion
    plugin: CredentialPlugin,
```

tests/hashivault_test.py
```diff
@@ -0,0 +1,229 @@
...
+    field_type: str,
+    expected_ids: list[str],
+) -> None:
+    """Verify plugin input fields/metadata are present with expected IDs."""
+    plugin_inputs = plugin.inputs  # type: ignore[attr-defined]
```
webknjaz: just fixe the type definition above
```suggestion
    plugin_inputs = plugin.inputs
```

tests/unit/credentials/hashivault_test.py
PabloHiro: You mean `TestDelineaImports`? This is a different credential plugin 😕

tests/unit/credentials/hashivault_test.py
PabloHiro: This is not a new test. This is a copy paste from credential_plugins_test.py. The only test I am adding is `test_plugin_input_ids`.

I am grouping the tests under `hashivault_test.py` so we have all tests for a plugin in a single file.

tests/unit/credentials/hashivault_test.py
PabloHiro: Same, this is not a new test. This is a copy paste from credential_plugins_test.py. The only test I am adding is `test_plugin_input_ids`.

I am grouping the tests under `hashivault_test.py` so we have all tests for a plugin in a single file.

I am not saying the feedback is not valid or it should not be addressed, I just think it is a separate issue to what I am trying to do here which is simply putting this test in a new file.

tests/unit/credentials/hashivault_test.py
PabloHiro: Same as above, just moving this test.

tests/unit/credentials/hashivault_test.py
PabloHiro: Same as above, just moving this test.

tests/unit/credentials/hashivault_test.py
PabloHiro: Same as above, just moving this test.

tests/hashivault_test.py
PabloHiro: If I do that, I get:
```
tests/hashivault_test.py:229:46:229:70: error: Value of type "object" is not
indexable  [index]
        actual_ids = [plugin['id'] for plugin in plugin_inputs[field_type]...
                                                 ^~~~~~~~~~~~~~~~~~~~~~~~~
```

Which I cannot ignore because the linting check for lines being too long complains x)

Why are we enforcing strict typing in a project that was not following good type hinting in the first place?

tests/unit/credentials/hashivault_test.py
PabloHiro: This one is:
```
  43:1     WPS118 Found too long name: test_hashivault_client_cert_auth_explicit_role > 45
           -> https://pyflak.es/WPS118
  def test_hashivault_client_cert_auth_explicit_role() -> None:
  ^
```

And again, this is an existing test that is coming from `credential_plugins_test.py`

tests/unit/credentials/hashivault_test.py
webknjaz: No, just rename the file. The runtime code is in `src/awx_plugins/credentials/hashivault.py` and so the corresponding tests could be in `tests/credentials/hashivault_test.py` — the same folder layout with a different prefix.

tests/hashivault_test.py
webknjaz: The old code has this disabled but there's no reason for new code to allow arbitrary violations by default.

If you accept the change I suggested in the signature, it'll improve things a bit: https://github.com/ansible/awx-plugins/pull/150/files#r2818571974. Of course, the index error fix would be more involved and so it's out of the scope here.

We can use justified suppressions if needed, with a comment explaining why.
If the  line is long, many syntax constructs do not require to be entirely within the same line and you should be able to multiline them.

tests/hashivault_test.py
```diff
@@ -0,0 +1,229 @@
...
+    actual_ids = [plugin['id'] for plugin in plugin_inputs[field_type]]
```
webknjaz: Something like this should help you:
```suggestion
    actual_ids = [
        plugin['id']
        # NOTE: `CredentialPlugin` aren't yet fully typed:
        for plugin in plugin_inputs[field_type]  # type: ignore[index]
    ]
```

PabloHiro: Done :+1:

tests/hashivault_test.py
PabloHiro: Done :+1:

tests/unit/credentials/hashivault_test.py
webknjaz: ~Doing this myself now.~ Decided to add `unit/` to the layout so that we could have a separation w/ integration and other test types.

.flake8
```diff
@@ -132,7 +132,7 @@ per-file-ignores =
   src/awx_plugins/credentials/plugins.py: B950,D100, D101, D103, D105, D107, D205, D400, LN001, WPS204, WPS229, WPS433, WPS440
   src/awx_plugins/credentials/tss.py: ANN003, ANN201, D100, D103, E712, WPS433, WPS440, WPS503
   src/awx_plugins/inventory/plugins.py: ANN001, ANN002, ANN003, ANN101, ANN102, ANN201, ANN202, ANN206, B950, C812, C819, D100, D101, D102, D205, D209, D400, D401, LN001, LN002, N801, WPS110, WPS111, WPS202, WPS210, WPS214, WPS301, WPS319, WPS324, WPS331, WPS336, WPS337, WPS338, WPS347, WPS421, WPS433, WPS450, WPS510, WPS529
-  tests/credential_plugins_test.py: ANN101, B017, C419, D100, D102, D103, D205, D209, D400, DAR, PT011, S105, WPS111, WPS117, WPS118, WPS202, WPS352, WPS421, WPS433, WPS507, WPS441
```
webknjaz, line 135: @PabloHiro I also dropped outdated suppressions that are either fixed with the move or would apply to the new module but not the old one. This is one of the places to clean up.

PR #151, opened 2026-02-11
fix: test test_azure_kv_invalid_env

PabloHiro:
## Summary by CodeRabbit

* **Tests**
  * Improved Azure Key Vault test reliability by implementing controlled mocking for credential unavailability scenarios, ensuring robust validation of error handling behavior without depending on external environment conditions.

### DIFF
--- a/tests/azure_kv_test.py
+++ b/tests/azure_kv_test.py
@@ -1,7 +1,10 @@
 """Tests Azure Key Vault credential plugin."""
 
+from unittest.mock import Mock
+
 import pytest
 
+from azure.identity import CredentialUnavailableError
 from azure.keyvault.secrets import (
     KeyVaultSecret,
     SecretClient,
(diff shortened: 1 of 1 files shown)


AlanCoding (approved):
seems like good CI maintenance

tests/azure_kv_test.py
```diff
@@ -1,7 +1,10 @@
 """Tests Azure Key Vault credential plugin."""
 
+from unittest.mock import Mock
```
webknjaz, line 3: @PabloHiro instead of importing `unittest.mock`, use the `mocker` fixture that's integrated into `pytest` via `pytest-mock` and is native to the framework in use.

tests/azure_kv_test.py
```diff
@@ -22,8 +25,23 @@ def get_secret(
...
+
+    When credentials are incomplete (e.g., empty client ID), the code falls
+    back to ManagedIdentityCredential. On a non-Azure VM, this raises
+    CredentialUnavailableError. Using a fake vault URL in tests would cause
+    a DNS error before the credential check, so we mock SecretClient to
+    simulate the expected CredentialUnavailableError.
+    """
+    mock_client = Mock()
+    mock_client.return_value.get_secret.side_effect = (
+        CredentialUnavailableError(
```
webknjaz, line 39: Instead of triggering the exception being directly handled, this should've triggered a situation when the library's internals would raise it. Otherwise, we're just testing own mocks and not how the real world would work.

PR #153, opened 2026-02-12
Add documentaiton on how to test

PabloHiro:


CONTRIBUTING.md
```diff
@@ -41,7 +41,7 @@ Open a GitHub issue to report bugs or request features.
 
 ## Governance
 
-The Ansible PROJECTNAME uses the following governance model:
+The Ansible awx-plugin uses the following governance model:
```
webknjaz: This is a separate change.

TESTING.md
webknjaz, line 1: This should be wired into the Sphinx docs, similar to contributing so it'll need a little infra change. But additionally, the content will need to be rewritten. Look into making it similar to https://github.com/ansible/pylibssh/blob/devel/docs/contributing/testing_guide.rst. I'll give more comprehensive advice tomorrow.

CONTRIBUTING.md
PabloHiro: Divided this into a separate commit

TESTING.md
PabloHiro: Moved it to the Sphinx docs, but I have not updated the contents. I am not familiar with sphinx syntax and I don't want to retest what works for a testing environment. I think it is better to have a simple working guide to non at all.

webknjaz: > sphinx syntax

Sphinx can use either RST or MD. For MD, you'll need to follow MyST syntax. Since I've originally integrated both parsers, your change has been picked up and renders w/o errors: https://awx-plugins-core--153.org.readthedocs.build/en/153/contributing/testing/.

I'll take a look to see what we need to improve syntax-wise and also in terms of the content.

docs/contributing/testing.md
```diff
@@ -0,0 +1,40 @@
...
+tox r -e py313
+```
+
+## Code Quality Checks
+
+### Pre-commit Hooks
+
+```bash
+# Install pre-commit
+pip install pre-commit
```
webknjaz: This is really unnecessary. Can be done optionally but doesn't deserve a confusing mention in the guide. Everythin is wired through tox and showing a dozen ways of doing the same is unhelpful. Those who'd want to use the underlying tooling directly will know but for those who don't we should show ways that differ from what the CI runs.

docs/contributing/testing.md
```diff
@@ -0,0 +1,40 @@
...
+- Python 3.11 due to compatibility issues
+- [tox](https://tox.wiki/) (recommended) or pip
+
+## Quick Start
+
+tox handles dependency installation and test execution automatically:
+
+```bash
+# Install tox
+pip install tox
```
webknjaz: Arguably, `pip` shouldn't be used directly outside of venvs. And since `tox` manages own venvs, this is really unncessary. We could mention `pipx` / `uvx` for such workflow apps and also link the docs.

docs/contributing/testing.md
```diff
@@ -0,0 +1,40 @@
+# Testing Guide
+
+This document explains how to install dependencies and run tests for awx-plugins.
+
+## Prerequisites
+
+- Python 3.11 due to compatibility issues
```
webknjaz: Nope.

docs/contributing/testing.md
```diff
@@ -0,0 +1,40 @@
...
+- [tox](https://tox.wiki/) (recommended) or pip
```
webknjaz: Not pip, no.

CONTRIBUTING.md
webknjaz: It's unrelated to testing and mustn't poison this PR. It's out of the scope.

PabloHiro: What should I do then? Open a separate PR?

docs/contributing/testing.md
PabloHiro: Removed :+1:

docs/contributing/testing.md
PabloHiro: Updated to Python 3

docs/contributing/testing.md
PabloHiro: Removed

docs/contributing/testing.md
PabloHiro: Removed

CONTRIBUTING.md
webknjaz: Yep.

docs/index.md
```diff
@@ -35,6 +35,7 @@ changelog
 
 Code Of Conduct <contributing/code_of_conduct>
 contributing/guidelines
+contributing/testing
 contributing/communication
 contributing/security
```
webknjaz: Let's order this similar to `ansible-pylibssh`:
```suggestion
contributing/communication
contributing/security
contributing/testing
```

docs/contributing/testing.md
```diff
@@ -0,0 +1,22 @@
...
+
+- Python 3
+- [tox](https://tox.wiki/)
+
+## Quick Start
+
+tox handles dependency installation and test execution automatically:
+
+```bash
+# Run the full test suite
```
webknjaz: It's rather unclear what "full" implies here. This runs the default toxenv under whatever `python` resolves to at the moment. It's probably more accurate to say `pytest`. I also like adding `-qq` to standard invocations lately — this hides output from `tox` and what's left is the `pytest` output (or other configured commands).

CONTRIBUTING.md
PabloHiro: Done: https://github.com/ansible/awx-plugins/pull/159

docs/index.md
PabloHiro: Done

docs/contributing/testing.md
PabloHiro: Done

PR #154, opened 2026-02-12
🧪 Make `flake8-logging-format` conditional

PabloHiro:
## Summary by CodeRabbit

* **Chores**
  * Updated development tool configuration to apply conditional dependency pins for specific Python version ranges and added explanatory compatibility notes.
  * Kept existing hook structure and other tool settings while adjusting only the relevant constraint lines.
* **Tests / CI**
  * Extended the lint job matrix to include the newest Python runner configuration so linting runs under the additional Python version.

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -379,6 +379,11 @@ jobs:
         - spellcheck-docs
         xfail:
         - false
+        include:
+        - python-version: 3.14
+          runner-vm-os: ubuntu-latest
+          toxenv: pre-commit
+          xfail: false
       fail-fast: false
 
     uses: tox-dev/workflow/.github/workflows/reusable-tox.yml@88781f74da62948ad3b87b803c95a6ad60c57028  # yamllint disable-line rule:line-length
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -177,8 +177,15 @@ repos:
     - flake8-docstrings ~= 1.7.0
     - flake8-length ~= 0.3.1
     - flake8-logging ~= 1.6.0
-    - flake8-logging-format ~= 2024.24.12
+    # NOTE: `flake8-logging-format` hasn't had updates since 2024 and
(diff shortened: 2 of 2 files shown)


webknjaz:
Plz explain where/how you saw this break/motivation.

We should probably go through #145, though.

webknjaz (dismissed):
Let's check which part of this is redundant post #145 merge and revert it. Then, let's get to the bottom of why you need the other changes.

PabloHiro:
Using python 3.12 or above locally without this change throws this:
```
$ pre-commit --run-all-files
...
flake8.exceptions.FailedToLoadPlugin: Flake8 failed to load plugin "flake8-logging-format" due to No module named 'pkg_resources'.
```

webknjaz:
Does that plugin have a new version? Any upstream issue recorded? Any more context?

webknjaz:
I think I'll be looking at https://adamj.eu/tech/2023/09/07/introducing-flake8-logging/ / https://docs.astral.sh/ruff/rules/logging-string-format/ to see how to migrate to a better compatible setup. Pinning the interpreter version is often problematic when it's not available in everyone's runtimes. Additionally, I'll see if the dep conditionals can be used more granularly. Stay tuned.

PabloHiro:
> Does that plugin have a new version? Any upstream issue recorded? Any more context?

@webknjaz  There is a lively discussion here: https://github.com/pypa/setuptools/issues/5174

webknjaz:
> I'll be looking at https://adamj.eu/tech/2023/09/07/introducing-flake8-logging/

Turns out, I've already integrated this. What's left is Ruff. I'll push fixes to this branch myself.

webknjaz:
> What's left is Ruff.

I realized that all its rules are enabled (except for the exclusions) so we're good here too.

webknjaz (approved):

webknjaz:
It's weird that the CI in this PR didn't hit https://github.com/m-burst/flake8-pytest-style/issues/361, but #156 does. This needs a follow-up deps bump.

PR #157, opened 2026-02-16
Add new credential types to support workload identity

PabloHiro:
AAP-64507

New credential plugins:
- HashiCorp Vault Secret Lookup (OIDC) - hashivault_kv_oidc_plugin
- HashiCorp Vault Signed SSH (OIDC) - hashivault_ssh_oidc_plugin

New authentication method:
- workload_identity_auth() - Handles JWT-based authentication using workload identity tokens

Refactored field definitions to individual field variables. This makes the definition of the credential types more readable. I also added unit tests.

Needs: https://github.com/ansible/awx-plugins/pull/150 

Screenshots:
image

## Summary by CodeRabbit

* **New Features**
  * Added OpenID Connect (OIDC) authentication support for HashiCorp Vault credentials.
  * Introduced workload identity (JWT) authentication flow for Vault access.
  * Added OIDC-enabled credential plugin variants for Vault KV and SSH, available as installable plugins.

* **Tests**
  * Added unit tests covering workload identity authentication and OIDC plugin inputs.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -69,6 +69,8 @@ name = "Ansible maintainers and contributors"
 conjur = "awx_plugins.credentials.conjur:conjur_plugin"
 hashivault_kv = "awx_plugins.credentials.hashivault:hashivault_kv_plugin"
 hashivault_ssh = "awx_plugins.credentials.hashivault:hashivault_ssh_plugin"
+hashivault-kv-oidc = "awx_plugins.credentials.hashivault:hashivault_kv_oidc_plugin"
+hashivault-ssh-oidc = "awx_plugins.credentials.hashivault:hashivault_ssh_oidc_plugin"
 azure_kv = "awx_plugins.credentials.azure_kv:azure_keyvault_plugin"
 aim = "awx_plugins.credentials.aim:aim_plugin"
 centrify_vault_kv = "awx_plugins.credentials.centrify_vault:centrify_plugin"
@@ -146,6 +148,14 @@ credentials-hashivault-kv = [
   "requests",
 ]
 credentials-hashivault-ssh = [
+  "awx_plugins.interfaces",
+  "requests",
+]
+credentials-hashivault-kv-oidc = [
+  "awx_plugins.interfaces",
+  "requests",
+]
+credentials-hashivault-ssh-oidc = [
   "awx_plugins.interfaces",
   "requests",
 ]
--- a/src/awx_plugins/credentials/hashivault.py
+++ b/src/awx_plugins/credentials/hashivault.py
@@ -1,7 +1,6 @@
 # FIXME: the following violations must be addressed gradually and unignored
 # mypy: disable-error-code="arg-type, no-untyped-call, no-untyped-def"
 
-import copy
 import os
 import pathlib
 import time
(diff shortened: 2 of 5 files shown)


src/awx_plugins/credentials/hashivault.py
```diff
@@ -300,6 +461,12 @@ def client_cert_auth(**kwargs):
     return {'name': kwargs.get('client_cert_role')}
 
 
+def workload_identity_auth(**kwargs):
+    """JWT representing a workload. Issued by an OIDC entity trusted by Vault."""
+    workload_identity_token = kwargs['workload_identity_token']
+    return {'role': kwargs.get('jwt_role'), 'jwt': workload_identity_token}
```
dleehr: For the token we use `kwargs['workload_identity_token']` (which may raise a `KeyError`), but for the role we use `kwargs.get('jwt_role')`, which won't raise a `KeyError` but may return `None`.

Any reason for the difference?

src/awx_plugins/credentials/hashivault.py
```diff
@@ -300,6 +461,12 @@ def client_cert_auth(**kwargs):
     return {'name': kwargs.get('client_cert_role')}
 
 
+def workload_identity_auth(**kwargs):
```
dleehr, line 466: Thanks for continuing the naming convention for workload identity 😁.

Nit: I think at this point (we're inside the plugin that authenticates with Vault), it makes sense to call this method JWT Authentication. The specific authentication we're using in Vault **is** [JWT Authentication](https://developer.hashicorp.com/vault/docs/auth/jwt#jwt-authentication).

dleehr (commented):
I'll focus the discussion in https://github.com/ansible/awx/pull/16286 but for now had a couple small pieces of feedback here.

src/awx_plugins/credentials/hashivault.py
```diff
@@ -16,246 +15,403 @@
...
+# OIDC-specific metadata
+job_template_metadata = {
+    'id': 'job_template',
+    'label': _('Job Template'),
+    'type': 'int',  # ID of the Job Template
+    'multiline': False,
+    'help_text': _(
+        'Job Template to use in order to generate a workload identity',
+    ),
+}
```
dleehr: Didn't catch this before, but what's the purpose of this `job_template` field in `metadata`?

I understand that we want users to provide a job template id for the test operation, but won't this also be considered/rendered in the normal use of the plugin? Maybe that's OK (and they'd just leave it blank), but it stands out a bit.

PabloHiro: Yeah... as you said, I know this would be consumed by the UI to render the fields displayed when one clicks on `Test` when creating the credentials. But I am not sure about its consequences in other places.

Maybe its premature to include it here, we can add conditional logic to display this field on the UI instead.

src/awx_plugins/credentials/hashivault.py
PabloHiro: Because the code here only gets called when we know that the key exists.

```
    elif kwargs.get('workload_identity_token'):
        token = method_auth(
            **kwargs,
            auth_param=workload_identity_auth(**kwargs),
        )
```
I will update it for consistency sake.

src/awx_plugins/credentials/hashivault.py
PabloHiro: Agree... but we have `kubernetes_auth` already, which also uses jwt. I think it is clearer to keep it separate based on features. Thoughts?

src/awx_plugins/credentials/hashivault.py
PabloHiro: Removed

src/awx_plugins/credentials/hashivault.py
dleehr: Very good point 👍

tests/unit/credentials/hashivault_test.py
tyraziel, line 1: Does it make sense to change the name of this to `hashivault_credential_plugin_test.py` ?

PabloHiro: There is another comment in other PR that suggests building a directory structure that imitates `src`. I am honestly gonna leave that up to the owners of the repo.

src/awx_plugins/credentials/hashivault.py
```diff
@@ -16,246 +15,388 @@
 from .plugin import CertFiles, CredentialPlugin, raise_for_status
 
 
-base_inputs = {
+# Base input fields
```
webknjaz, line 19: Let's imitate something like this for anything this patch touches: https://github.com/ansible/awx-plugins/blob/b1e1ee853c5744cf65d906da458b003fa66059a3/src/awx_plugins/credentials/github_app.py#L37-L64. For old stuff that's been migrated as is initially, it ways okay'ish to keep it highly dynamic but as we change or add things, they should be typed.

src/awx_plugins/credentials/hashivault.py
```diff
@@ -1,7 +1,6 @@
 # FIXME: the following violations must be addressed gradually and unignored
 # mypy: disable-error-code="arg-type, no-untyped-call, no-untyped-def"
```
webknjaz, line 2: As you apply typed dicts, check if we can drop any of these suppressions.

tests/unit/credentials/hashivault_test.py
webknjaz: Rebasing on top of #150 will move the test module introduced there. Renaming again is likely redundant.

src/awx_plugins/credentials/hashivault.py
PabloHiro: @webknjaz  I moved these types to a new file named `types.py` and updated all plugins to use these types.

src/awx_plugins/credentials/hashivault.py
PabloHiro: I tried to get rid of these locally but I did not succeed. I would end up needing to ignore other errors just to get rid of `no-untyped-def`.

webknjaz: Okay, then. Marking this as resolved.

pyproject.toml
```diff
@@ -69,6 +69,8 @@ name = "Ansible maintainers and contributors"
 conjur = "awx_plugins.credentials.conjur:conjur_plugin"
 hashivault_kv = "awx_plugins.credentials.hashivault:hashivault_kv_plugin"
 hashivault_ssh = "awx_plugins.credentials.hashivault:hashivault_ssh_plugin"
+hashivault_kv_oidc = "awx_plugins.credentials.hashivault:hashivault_kv_oidc_plugin"
+hashivault_ssh_oidc = "awx_plugins.credentials.hashivault:hashivault_ssh_oidc_plugin"
```
webknjaz: I recommend using kebab-case for the entries so they aren't confusable with importables. Been meaning to have this style from the beginning.
```suggestion
hashivault-kv-oidc = "awx_plugins.credentials.hashivault:hashivault_kv_oidc_plugin"
hashivault-ssh-oidc = "awx_plugins.credentials.hashivault:hashivault_ssh_oidc_plugin"
```

src/awx_plugins/credentials/types.py
webknjaz, line 1: Let's rename this module to `_types.py` since it hasn't been well-thought as a public API. And it's likely that we'll want to host the types in the interfaces repo later.

src/awx_plugins/credentials/types.py
```diff
@@ -0,0 +1,37 @@
+"""Shared TypedDict definitions for credential plugin schemes."""
+
+from typing import NotRequired, TypedDict
```
webknjaz: Plz change all new uses of `typing` to use the following import style:
```suggestion
import typing as _t
```

I'm going to integrate a flake8 plugin to enforce this later on.

src/awx_plugins/credentials/types.py
```diff
@@ -0,0 +1,37 @@
+"""Shared TypedDict definitions for credential plugin schemes."""
```
webknjaz: Let's not reference implementation detail literals in docstring titles: https://awx-plugins-core--157.org.readthedocs.build/en/157/pkg/awx_plugins.credentials.types/
```suggestion
"""Shared interface type definitions for credential plugin schemas."""
```

src/awx_plugins/credentials/tss.py
webknjaz, line 1: This became a rather big refactoring — let's move type upgrades/relocation into a separate PR.

tests/unit/credentials/hashivault_test.py
webknjaz: Assuming there's no further action items here, I'm marking this thread as resolved.

tests/unit/credentials/hashivault_test.py
```diff
@@ -75,6 +81,17 @@ def test_hashivault_userpass_auth() -> None:
     assert res == expected_res
 
 
+def test_hashivault_workload_identity_auth() -> None:
+    """Test workload_identity_auth."""
```
webknjaz: Make the docstring describe the expectations accurately. It's unclear what the test tests.

src/awx_plugins/credentials/types.py
PabloHiro: This was addressed in: https://github.com/ansible/awx-plugins/pull/161

src/awx_plugins/credentials/types.py
PabloHiro: This was addressed in: https://github.com/ansible/awx-plugins/pull/161

src/awx_plugins/credentials/types.py
PabloHiro: This was addressed in: https://github.com/ansible/awx-plugins/pull/161

src/awx_plugins/credentials/tss.py
PabloHiro: This was addressed in: https://github.com/ansible/awx-plugins/pull/161

tests/unit/credentials/hashivault_test.py
PabloHiro: Done

src/awx_plugins/credentials/hashivault.py
```diff
@@ -475,3 +630,15 @@ def ssh_backend(**kwargs):
...
+hashivault_kv_oidc_plugin = CredentialPlugin(
+    'HashiCorp Vault Secret Lookup (OIDC)',
+    inputs=hashi_kv_oidc_inputs,
+    backend=kv_backend,
+)
+
+hashivault_ssh_oidc_plugin = CredentialPlugin(
+    'HashiCorp Vault Signed SSH (OIDC)',
+    inputs=hashi_ssh_oidc_inputs,
+    backend=ssh_backend,
```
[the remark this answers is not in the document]
PabloHiro: @webknjaz I am not too familiar with pyproject.toml, is the comment above accurate? Please let me know

src/awx_plugins/credentials/hashivault.py
PabloHiro: @dleehr closing this as I assume you are okey with the current name. Let me know if it is not the case.

src/awx_plugins/credentials/hashivault.py
webknjaz: Marking this as resolved.

src/awx_plugins/credentials/hashivault.py
webknjaz: Personally, I'm unhappy with function names being nouns.. A thing that does something isn't its result. Callables are supposed to be verbs. But then again, it's not a refactoring effort. Maybe one day.

src/awx_plugins/credentials/hashivault.py
```diff
@@ -300,6 +449,12 @@ def client_cert_auth(**kwargs):
     return {'name': kwargs.get('client_cert_role')}
 
 
+def workload_identity_auth(**kwargs):
+    """JWT representing a workload. Issued by an OIDC entity trusted by Vault."""
```
webknjaz, line 467: @PabloHiro FYI here's typical pythonic docstring convensions: https://peps.python.org/pep-0257/. A short one-sentence title/summary, followed by a paragraph separated with an empty line. Note that since callables do something, the summary line is similar in style to Git commits: starts with a verb, uses imperative mood, is a bit high-level.

src/awx_plugins/credentials/hashivault.py
```diff
@@ -300,6 +449,12 @@ def client_cert_auth(**kwargs):
...
+    workload_identity_token = kwargs.get('workload_identity_token')
+    return {'role': kwargs.get('jwt_role'), 'jwt': workload_identity_token}
```
webknjaz, line 469: *(not renaming the callable, although it'd be nice)*

Let's try getting rid of the dynamic stuff as much as possible while preserving the signature:
```suggestion
def workload_identity_auth(
    *,
    jwt_role: str,
    workload_identity_token: str,
    **_discarded_kwargs: object,
) -> dict[str, str]:
    """Compose a JWT object representing a workload.
    
    The embedded token is assumed to have been issued by an OIDC entity trusted by Vault.
    """
    return {'role': jwt_role, 'jwt': workload_identity_token}
```

(the return type could be better feel free to iterate now or later)

---

P.S. It's probably possible to have a `jwt_role: str | None = None` but I'm not sure if we'd want to bother — are there cases when it's not passed? We could just add a check before calling this function, if needed.

---

You may want to try this for the discarded args. I haven't tested it locally.

```suggestion
def workload_identity_auth(
    *,
    jwt_role: str,
    workload_identity_token: str,
    **_discarded_kwargs: dict[_t.Never, _t.Never],
) -> dict[str, str]:
    """Compose a JWT object representing a workload.
    
    The embedded token is assumed to have been issued by an OIDC entity trusted by Vault.
    """
    return {'role': jwt_role, 'jwt': workload_identity_token}
```

src/awx_plugins/credentials/hashivault.py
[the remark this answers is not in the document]
webknjaz: Sounds about right. Your changes to that config look reasonable.

src/awx_plugins/credentials/hashivault.py
```diff
@@ -475,3 +630,15 @@ def ssh_backend(**kwargs):
     inputs=hashi_ssh_inputs,
     backend=ssh_backend,
 )
+
+hashivault_kv_oidc_plugin = CredentialPlugin(
+    'HashiCorp Vault Secret Lookup (OIDC)',
```
webknjaz, line 649: Do we need to append something in parens for the old plugins too?

src/awx_plugins/credentials/hashivault.py
```diff
@@ -13,249 +12,394 @@
...
+        secret_backend_metadata,
+        secret_path_metadata,
+        auth_path_metadata,
+        secret_key_metadata,
+        secret_version_metadata,
     ],
     'required': [
         'url',
         'secret_path',
+        'api_version',
```
webknjaz: This and other places ended up with copies of field identifiers still duplicated but now more disconnected and spread out across the module. For this reason, I think we should rewrite all the entries to reference the same vars that you've moved out and access the `'id'` key like so:
```suggestion
        api_version_field['id'],
```

*(here and everywhere else)*

tests/unit/credentials/hashivault_test.py
```diff
@@ -77,6 +77,17 @@ def test_hashivault_userpass_auth() -> None:
...
 
+def test_hashivault_workload_identity_auth() -> None:
+    """Test ``workload_identity_auth()`` returns the token."""
+    kwargs = {
+        'workload_identity_token': 'the_jwt_token',
+        'jwt_role': 'the_jwt_role',
+    }
+    expected_res = {'role': 'the_jwt_role', 'jwt': 'the_jwt_token'}
+    res = hashivault.workload_identity_auth(**kwargs)  # type: ignore[no-untyped-call]
+    assert res == expected_res
```
webknjaz: I've made a suggestion in the respective module to improve typing. Additionally, here we can just pass the args explicitly and drop the proxy object with a generic meaningless name that's being unpacked creating unnecessary indirection in the test. Finally, the actual interface accepts an arbitrary number of args which aren't tested here. So we can excercise that too.

```suggestion
@pytest.mark.parametrize(
    'arbitrary_kwargs',
    ({}, {'utter': 'nonsense'}),
    ids=('no-additional-keyword-args', 'with-additional-keyword-arg'),
)
def test_hashivault_workload_identity_auth(arbitrary_kwargs: dict[str, str]) -> None:
    """Test ``workload_identity_auth()`` returns the token."""
    sentinel_role_value = 'the_jwt_role'
    sentinel_jwt_value = 'the_jwt_token'
    expected_jwt_object = {'role': sentinel_role_value, 'jwt': sentinel_jwt_value}

    computed_jwt_object = hashivault.workload_identity_auth(
        workload_identity_token=sentinel_jwt_value,
        jwt_role=sentinel_role_value,
        **arbitrary_kwargs,
    )

    assert computed_jwt_object == expected_jwt_object
```

webknjaz: Looking more into this, it would've made sense to update `test_hashivault_handle_auth_token` with parametrization that would cover calling `workload_identity_auth()` among other cases. You could use `mocker.spy()` to check how this underlying method is being called w/o distrupting the control flow. There's really no need for a separate test that just checks a simple function like `hashivault.workload_identity_auth()` that does absolutely nothing short of returning reshuffled data when its caller is simple enough and testing it would create better / more complete coverage and embed more realistic / integrated scenarios into the test suite.

tests/unit/credentials/hashivault_test.py
webknjaz, line 1: Just remembered... Could you also add a smoke test (parametrize entry) to https://github.com/ansible/awx-plugins/blob/devel/tests/importable_test.py ?

tests/unit/credentials/hashivault_test.py
```diff
@@ -142,6 +153,27 @@ def test_hashivault_handle_auth_client_cert(mocker: MockerFixture) -> None:
     assert token == 'the_token'
 
 
+def test_hashivault_handle_auth_workload_identity(
+    mocker: MockerFixture,
+) -> None:
+    """Test handle auth with workload identity auth."""
+    kwargs = {
```
webknjaz: Let's make the var names less arbitrary:
```suggestion
    workload_id_kwargs = {
```

tests/unit/credentials/hashivault_test.py
```diff
@@ -142,6 +153,27 @@ def test_hashivault_handle_auth_client_cert(mocker: MockerFixture) -> None:
...
+        'jwt_role': 'the_jwt_role',
+        'default_auth_path': 'jwt',
+        'url': 'https://vault.example.com',
+    }
+    auth_params = {
+        'role': 'the_jwt_role',
+        'jwt': 'the_jwt_token',
+    }
+    method_mock = mocker.patch.object(hashivault, 'method_auth')
+    method_mock.return_value = 'the_token'
```
webknjaz:
```suggestion
    method_mock = mocker.patch.object(hashivault, 'method_auth', autospec=True, return_value='the_token')
```

src/awx_plugins/credentials/hashivault.py
PabloHiro: No, the name was chosen by the product manager leading this feature. Old names stay the same.

src/awx_plugins/credentials/hashivault.py
PabloHiro: I am not gonna implement this change. This deviates from how the other methods are defined and falls into the refactor category. If this gets refactored in the future, a better interface than a bunch of kwargs can be thought of, and the AWX code can be updated accordingly.

I think doing the above for just this function would confuse whoever is trying to modify this code.

tests/unit/credentials/hashivault_test.py
PabloHiro: Done

src/awx_plugins/credentials/tss.py
webknjaz: Marking as resolved.

src/awx_plugins/credentials/types.py
webknjaz: Marking as resolved.

tests/unit/credentials/hashivault_test.py
webknjaz: Marking as resolved.

src/awx_plugins/credentials/types.py
webknjaz: Marking as resolved.

src/awx_plugins/credentials/types.py
webknjaz: Marking as resolved.

pyproject.toml
webknjaz: Marking as resolved.

src/awx_plugins/credentials/hashivault.py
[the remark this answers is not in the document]
webknjaz: Marking as resolved.

src/awx_plugins/credentials/hashivault.py
webknjaz: I disagree: this is a newly added function and copying a terrible earlier practice only makes things worse. The suggested style does not change the invocation mechanics, it only makes it readable and possible to reason about. I'm unable to see how making the function look cleaner is more confusing than copying dynamically referenced dynamic stuff returned dynamically.

But if you want the exact same behavior, here's 

```suggestion
def workload_identity_auth(
    *,
    jwt_role: str | None = None,
    workload_identity_token: str | None = None,
    **_discarded_kwargs: dict[_t.Never, _t.Never],
) -> dict[str, str | None]:
    """Compose a JWT object representing a workload.
    
    The embedded token is assumed to have been issued by an OIDC entity trusted by Vault.
    """
    return {'role': jwt_role, 'jwt': workload_identity_token}
```

This variant does not have a single behavior difference with your variant and doesn't refactor anything pre-existing (unlike a bunch of changes to the schema earlier in this module).

tests/unit/credentials/hashivault_test.py
webknjaz: Marking as resolved.

tests/unit/credentials/hashivault_test.py
```diff
@@ -77,6 +77,32 @@ def test_hashivault_userpass_auth() -> None:
...
+    """Test ``workload_identity_auth()`` returns the token."""
+    sentinel_role_value = 'the_jwt_role'
+    sentinel_jwt_value = 'the_jwt_token'
+    expected_jwt_object = {
+        'role': sentinel_role_value,
+        'jwt': sentinel_jwt_value,
+    }
+
+    auth = hashivault.workload_identity_auth
+    computed_jwt_object = auth(  # type: ignore[no-untyped-call]
```
webknjaz, line 97: Let's fix the typing instead: ignores are fine if justified/impossible to fix but for new code, I'd expect most violations to be avoided instead of suppressed.

src/awx_plugins/credentials/hashivault.py
dleehr: @webknjaz thank you for all of your feedback. I recognize there are opportunities to improve this codebase and embrace best practices. You've provided a lot of expertise towards those practices, and @PabloHiro has graciously accepted that feedback and been very responsive in his efforts.

That said, we continue to focus a lot of communication and then effort on changes that are not getting us to alignment with stakeholders on code architecture or functionality. That puts our ability to deliver this work at risk. This thread for example is debating a two-line function that is nearly trivial and clearly adopts an existing pattern.

Can we please make sure the focus is on the functionality we need to deliver?

webknjaz: FWIW, my approval isn't technically needed as long as I don't post a blocking review which I didn't: if somebody from the team owning the project wants to merge this — that's their decision, not mine.

---

As for my personal courtesy review — it's not up to the level of quality I'm seeking, which is what my feedback is about. I've skipped asking to separate unrelated refactoring at this time but beyond that I can't make it any easier. My interest here is in preserving good state of the project despite it not being my responsibility. I'm not blocking PRs as long as they don't trash the project but I'm also not posting approving reviews lightly.

webknjaz:
> [!important]
> The pre-commit jobs are failing due to an ongoing coveralls outage: https://status.coveralls.io/.
> It's likely that merging this PR will require an override (the rules bypass checkbox).
> So said job failures can be ignored as long as they fail in the coveralls upload step and not earlier. This has to be checked manually prior to merging.
> The linkcheck job failures have the same unrelying issue — it attempts checking links to coveralls/badges and gets an HTTP 500.

UPD: I've filed a tracking issue #164 and a temporary workaround #165.

webknjaz:
Alright. I've handled the CI stuff and clicked "Rebase" so that it's absorbed into this PR. If anybody from the controller team wants to merge this — they'll need to review the patch and approve it.

src/awx_plugins/credentials/hashivault.py
```diff
@@ -13,249 +12,394 @@
...
+    'id': 'jwt_aud',
+    'label': _('JWT Audience'),
+    'type': 'string',
+    'multiline': False,
+    'help_text': _(
+        'Identifies this Vault instance as the intended recipient of JWTs. '
+        'This value must match the bound_audiences configuration in your '
+        'Vault JWT auth method. Examples: hashicorp-vault-prod-01, '
+        'https://vault.example.com, or vault-production.',
+    ),
```
dleehr, line 189: Since this plugin is for HashiCorp Vault, the intended recipient will always be a HashiCorp Vault instance. I think it's fair to provide a default like `vault.workload.identity`, but yes it should be configurable.

webknjaz: You can post a comment with suggested changes (step 8 @ https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/reviewing-proposed-changes-in-a-pull-request#starting-a-review) to show exactly what you request.

webknjaz:
Since the coveralls outage is over sooner than anticipated, I've restored the integration and clicking "rebase" again..

AlanCoding:
I don't like the idea of hacking in an additional kwarg, so I would like for the new workload identity JWT to be formally declared as a field, which takes some novel changes to the overall structure. Here is my initial AI spitball of that idea:

https://github.com/ansible/awx-plugins/commit/d6d4eaab026361b2412e448342857b8d65bbf5c5

Right now we still need to check that we're all aligned on the high-level idea. I do hope this is minimal enough of an adjustment to still get it implemented relatively speedily.

PabloHiro:
> I don't like the idea of hacking in an additional kwarg, so I would like for the new workload identity JWT to be formally declared as a field, which takes some novel changes to the overall structure. Here is my initial AI spitball of that idea:
> 
> [d6d4eaa](https://github.com/ansible/awx-plugins/commit/d6d4eaab026361b2412e448342857b8d65bbf5c5)
> 
> Right now we still need to check that we're all aligned on the high-level idea. I do hope this is minimal enough of an adjustment to still get it implemented relatively speedily.

@AlanCoding updated with your commit, had to add another commit to address the linting

tests/unit/credentials/hashivault_test.py
```diff
@@ -231,3 +336,55 @@ def test_plugin_input_ids(
...
+    assert field['secret'] is True
+
+
+def test_workload_identity_token_field_not_req() -> None:
+    """Verify workload_identity_token is not in the required list for OIDC plugins."""
+    for inputs in (
+        hashivault.hashi_kv_oidc_inputs,
+        hashivault.hashi_ssh_oidc_inputs,
+    ):
+        assert 'workload_identity_token' not in inputs['required']
```
webknjaz: I'd represent this as two tests via `@pytest.mark.parametrize` as well:
```suggestion
@pytest.mark.parametrize(
    'plugin_inputs',
    (hashivault.hashi_kv_oidc_inputs, hashivault.hashi_ssh_oidc_inputs),
    ids=('kv-oidc', 'ssh-oidc'),
)
def test_workload_identity_token_field_not_req(
    plugin_inputs: _types.PluginInputs,
) -> None:
    """Verify ``workload_identity_token`` is not in the required list for OIDC plugins."""
    assert 'workload_identity_token' not in plugin_inputs['required']
```

(other than this, the CI is green atm — I've restarted the flaky job)

src/awx_plugins/credentials/hashivault.py
```diff
@@ -13,249 +12,408 @@
...
+        'job can access. Example: controller-automation, aap-production-jobs. '
+        'This must match a role created in Vault at auth/jwt/role/<role-name>.',
+    ),
+}
+
+jwt_audience_field: _types.FieldDict = {
+    'id': 'jwt_aud',
+    'label': _('JWT Audience'),
+    'type': 'string',
+    'multiline': False,
```
dleehr, line 183:
```
    'multiline': False,
    'default': 'vault.workload.identity',
```

I mentioned this suggestion in https://github.com/ansible/awx-plugins/pull/157/changes#r2865908186 but I'm reconsidering.

For a simple user experience I suggested having a default value for JWT Audience that's a simple string like `vault.workload.identity`. That's a pattern I see in [Terraform's Vault integration](https://developer.hashicorp.com/terraform/enterprise/workspaces/dynamic-provider-credentials/vault-configuration#optional-environment-variables). An audience like that basically says "This token is for **a** vault server, not AWS or Azure" which is true and one less thing for users to decide. 

After revisiting the standard https://www.rfc-editor.org/rfc/rfc7519#section-4.1.3 and doing some basic research, I think I agree with the strategy of **not** providing a default and guiding users to provide something specific to their specific Vault instance.

src/awx_plugins/credentials/hashivault.py
dleehr: 👍

src/awx_plugins/credentials/hashivault.py
```diff
@@ -13,249 +12,408 @@
...
+        'https://vault.example.com, or vault-production.',
+    ),
+}
+
+workload_identity_token_field: _types.FieldDict = {
+    'id': 'workload_identity_token',
+    'label': _('Workload Identity Token'),
+    'type': 'string',
+    'secret': True,
+    'internal': True,
```
dleehr, line 197: I followed the discussion about this field, no objection on the code here.

Two follow-up questions:
1. Does this `internal` property  on the Field require a change to https://github.com/ansible/awx_plugins.interfaces (perhaps [credential.pyi]( https://github.com/ansible/awx_plugins.interfaces/blob/devel/_type_stubs/awx/main/models/credential.pyi))?
2. When we merge this, how is the user experience impacted? Is this functional immediately? Or do we need API/UI changes to make this functional?

dleehr (approved):
LGTM.

Seems like a lot of refactoring/reorganization, and I think the only new pattern here is the addition of the `internal` property used by the new `workload_identity_token` fields.

I left a follow up question around the impacts of that, but what's here makes sense.

src/awx_plugins/credentials/hashivault.py
dleehr: Ah I see it's "not required"

webknjaz: 1 - Honestly, I wish many things were shaped as public interfaces and moved in there but that would need a proper effort. If people tested it and it didn't explode, then unlikely.
2 - Dunno about the awx side but this repo will need a release cut so that it's got a public version number. Lucky for us, I made the release automation single-click and it can be triggered anytime with the release artifacts hitting PyPI in just over 5 minutes (if the cooldown timer is not skipped, it'll have extra delay).

webknjaz: > Ah I see it's "not required"

That's on the type checking level. Runtime has absolutely no checks for this and many places outside `awx-plugins-core` and `awx_plugins.interfaces` don't have any typing whatsoever. Though, in these moved-out libs I've been trying to set up better linting and testing standards because it's easier due to their size (compared to the huge awx code base).

PabloHiro: 1 -> Yes. The change is present in 
2 -> It looks like this in the UI:
image

Notice that the `Workload Identity Token` field is shown by default. UI needs to be updated to filter out internal fields. A user can create them, and consume them as input sources in other credentials, but they do not work when running jobs. That again requires: https://github.com/ansible/awx/pull/16286

src/awx_plugins/credentials/hashivault.py
webknjaz: Yep, having mandatory fields is often a better API for non-userfacing stuff. `python -Im this | head -n-7 | tail -n1`

src/awx_plugins/credentials/hashivault.py
dleehr: Yep I noticed https://github.com/ansible/awx/pull/16286/changes#diff-a4bfa7b8ec60995660f9bd73750af496eb70e890974c1d636f2265c6289ffc94R431-R433 in the other PR that causes the field to be ignored. Thanks!

```
            # Prevent users from providing values for internally resolved fields
            if field.get('internal', False):
                continue
```

davemulford (approved):
Looks great!

AlanCoding (approved):

src/awx_plugins/credentials/hashivault.py
```diff
@@ -272,6 +430,11 @@ def handle_auth(**kwargs):
...
     ):
         token = method_auth(**kwargs, auth_param=client_cert_auth(**kwargs))
+    elif kwargs.get('workload_identity_token'):
+        token = method_auth(
+            **kwargs,
+            auth_param=workload_identity_auth(**kwargs),
+        )
     else:
         raise Exception(
             'Token, Username/Password, AppRole, Kubernetes, or TLS authentication parameters must be set',
```
arrestle, line 440:
```suggestion
elif kwargs.get('jwt_role') and not kwargs.get('workload_identity_token'): # if it's a vault token
    raise Exception(
        'Workload Identity requires workload_identity_token. '
        'Ensure this is an OIDC credential type and FEATURE_OIDC_WORKLOAD_IDENTITY_ENABLED is set.'
    )
else:
    raise Exception(
        'Token, Username/Password, AppRole, Kubernetes, TLS, or Workload Identity authentication parameters must be set',
    )
```

tests/unit/credentials/hashivault_test.py
```diff
@@ -231,3 +336,58 @@ def test_plugin_input_ids(
     plugin_inputs = plugin.inputs
     actual_ids = [plugin['id'] for plugin in plugin_inputs[field_type]]
```
arrestle, line 337:
```suggestion
+ actual_ids = [field['id'] for field in plugin_inputs[field_type]]
```
Use a different loop variable to avoid shadowing the plugin parameter:

arrestle (changes requested):
ANSTRAT-1558 calls for revoking the Vault client token after secrets are retrieved. Can you note whether that's in scope for this PR or deferred? If deferred, a brief mention or follow-up ticket would help set expectations.

Consider adding a targeted hint when jwt_role is set but workload_identity_token is missing:

elif kwargs.get('jwt_role') and not kwargs.get('workload_identity_token'):
    raise Exception(
        'Workload Identity requires workload_identity_token (injected by Controller). '
        'Ensure this is an OIDC credential type and FEATURE_OIDC_WORKLOAD_IDENTITY_ENABLED is set.'
    )

arrestle (approved):
ANSTRAT-1558 calls for revoking the Vault client token after secrets are retrieved. Can you note whether that's in scope for this PR or deferred? If deferred, a brief mention or follow-up ticket would help set expectations.

tests/unit/credentials/hashivault_test.py
tyraziel: Will follow up with a new JIRA card.

dleehr:
> ANSTRAT-1558 calls for revoking the Vault client token after secrets are retrieved. Can you note whether that's in scope for this PR or deferred? If deferred, a brief mention or follow-up ticket would help set expectations.

That requirement is covered in AAP-64519. This PR delivers part of AAP-64507, not the entire ANSTRAT.

src/awx_plugins/credentials/hashivault.py
webknjaz: @arrestle things like feature flags shouldn't leak into this lib. Also, when raising more exceptions, it's best to avoid the base class and raise something specific so that the calling side would be able to differentiate it (and might re-raise with a message that makes sense in its context outside this library).

tests/unit/credentials/hashivault_test.py
webknjaz: It's also a good idea b/c the data struct probably represents a different abstraction..

webknjaz:
@PabloHiro post-merge follow-ups needed in awx:
* bump the env pins for this lib to `0.1.0a1`
* list all the extras in places where the direct deps are listed

PR #161, opened 2026-02-23
Add common types for all plugins

PabloHiro:
## Summary by CodeRabbit

* **Refactor**
  * Added centralized, explicit type definitions for credential plugin inputs and applied annotations across credential plugins to improve static checks and editor support.

* **Tests**
  * Tightened test input typing and simplified test logic for credential input validation.

* **Chores**
  * Updated linting configuration to accommodate the new typing setup.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -179,6 +179,9 @@ rst-roles =
   event,
 
 # wemake-python-styleguide
+# WPS111:
+# `_t` will be enforced by `flake8-typing-as-t`
+allowed-domain-names = _t
 i-control-code = false
 # WPS226:
 max-string-usages = 85
--- a/src/awx_plugins/credentials/_types.py
+++ b/src/awx_plugins/credentials/_types.py
@@ -1,0 +1,37 @@
+"""Shared interface type definitions for credential plugin schemes."""
+
+import typing as _t
+
+
+class FieldDict(_t.TypedDict):
+    """A single UI field schema."""
+
+    id: str
+    label: str
+    type: _t.NotRequired[str]
+    format: _t.NotRequired[str]
+    secret: _t.NotRequired[bool]
+    multiline: _t.NotRequired[bool]
+    help_text: _t.NotRequired[str]
+    default: _t.NotRequired[str | bool]
+    choices: _t.NotRequired[list[str]]
+
+
+class MetadataDict(_t.TypedDict):
+    """Schema for input metadata."""
+
+    id: str
+    label: _t.NotRequired[str]
+    type: _t.NotRequired[str]
+    help_text: _t.NotRequired[str]
+    multiline: _t.NotRequired[bool]
+    default: _t.NotRequired[str | bool]
+    choices: _t.NotRequired[list[str]]
+
+
+class PluginInputs(_t.TypedDict):
+    """Schema for a collection of plugin input fields."""
+
+    fields: list[FieldDict]
+    metadata: list[MetadataDict]
+    required: list[str]
--- a/src/awx_plugins/credentials/plugin.py
+++ b/src/awx_plugins/credentials/plugin.py
@@ -7,12 +7,14 @@ import typing
 
 from requests.exceptions import HTTPError
 
+from . import _types
+
 
 class CredentialPlugin(typing.NamedTuple):
     """Schema for credential plugins."""
 
     name: str
-    inputs: object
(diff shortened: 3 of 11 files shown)


src/awx_plugins/credentials/types.py
webknjaz, line 1: I asked to move this to a private name in the other PR. The same applies here. `_types.py`.

tests/unit/credentials/hashivault_test.py
```diff
@@ -1,5 +1,7 @@
 """Tests for HashiCorp Vault credential plugins."""
 
+from typing import Literal
```
webknjaz:
```suggestion
import typing as _t
```

src/awx_plugins/credentials/plugin.py
```diff
@@ -7,12 +7,14 @@
 
 from requests.exceptions import HTTPError
 
+from .types import PluginInputs
```
webknjaz: Use the following style here to avoid adding public names into the global namespace:
```suggestion
from . import _types
```

webknjaz (commented):
Here's a few pointers I copied from #157. Also, add `allowed-domain-names = _t` at the end of `.flake8` with the rule number in the comment.

.flake8
webknjaz, line 1: I asked to move this to the end because there are other settings of the same plugin that are grouped there. This setting does not belong among stdlib options.

PabloHiro: Done

src/awx_plugins/credentials/plugin.py
PabloHiro: Done

tests/unit/credentials/hashivault_test.py
PabloHiro: Done

src/awx_plugins/credentials/types.py
PabloHiro: Done

.flake8
```diff
@@ -184,3 +184,6 @@ i-control-code = false
 max-string-usages = 85
 show-violation-links = true
 show-source = true
+
+# Exclude these names from other rules
```
webknjaz: It's a part of `wemake-python-styleguide` above. I'll merge them.

tests/unit/credentials/hashivault_test.py
webknjaz: I'll do this.

src/awx_plugins/credentials/_types.py
```diff
@@ -0,0 +1,37 @@
...
+
+import typing as _t
+
+
+class FieldDict(_t.TypedDict):
+    """A single UI field schema."""
+
+    id: str
+    label: str
+    type: _t.NotRequired[str]
```
webknjaz, line 11: @PabloHiro I think this is almost ready but could you comment on the changes you've made to the old types before we merge? I see that some are indeed moved but fileds like this are wrapped additionally.

PabloHiro: When trying to use this new type for all the plugins, it seems like these fields seem to be optional, as some fields in some plugins do not populate `type` and they work anyway...

webknjaz: @PabloHiro okay.. Any other notable differences with the previous state?

src/awx_plugins/credentials/types.py
webknjaz: Marking as resolved.

src/awx_plugins/credentials/plugin.py
webknjaz: Marking as resolved.

src/awx_plugins/credentials/_types.py
PabloHiro: What draw my attention was that `MetadataDict` used both `default` and `choices`, I am not familiar with every credential type, so I simply added support for those and assumed the definition of the field in the source code is correct.

But apart from that all I did was marking most fields as `NotRequired`

.flake8
webknjaz: Marking this as resolved.

src/awx_plugins/credentials/_types.py
webknjaz: Alright, so this would be something to explore separately as we get closer to making the types public.

webknjaz (approved):

PR #175, opened 2026-03-11
Fix #21: Prevent pytest from re-running full suite on coverage failure

### LINKED ISSUE #21: re-run failed tests triggers on failed coverage
https://github.com/ansible/awx-plugins/actions/runs/10597353901/job/29367631346?pr=17

`FAIL Required test coverage of 100% not reached. Total coverage: 40.78%`

It looks like the re-run failed tests with higher verbosity triggers re-running ALL tests when the coverage check fails. In the case above all the tests passed both runs.

TheDevCodingKen:
Fixes #21

### Summary
When `pytest-cov` triggers a failure due to the coverage threshold not being met, the `pytest --lf` command in the `tox-rerun-posargs` list evaluates an empty failure cache. By default, `pytest` responds to an empty `--lf` cache by re-running the entire test suite, which leads to inefficient use of CI resources.

This PR adds the `--last-failed-no-failures none` flag to the `ci-cd.yml` workflow. This overrides the default behavior, instructing `pytest` to fail fast and skip test execution if there are no actual functional test failures in the cache.

### Local Verification
Reproduced the issue locally by temporarily raising the `fail_under` threshold in `.coveragerc` to force a coverage failure, then executing the rerun logic.

**Before (Simulating coverage drop):**
```bash
.tox/py/bin/python -m pytest --no-cov --lf
```

**After (With flag applied):**
```bash
.tox/py/bin/python -m pytest --no-cov --lf --last-failed-no-failures none
```

**Result:** 
Pytest correctly identified no test failures and deselected the suite.
```
run-last-failure: no previously failed tests, deselecting all items.
========================= 132 deselected in 0.91s ==========================
```

## Summary by CodeRabbit

* **Chores**
  * Updated CI/CD configuration to improve test rerun behavior so retrying tests won’t error when there are no prior failures.

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -540,6 +540,7 @@ jobs:
         --no-cov
         -vvvvv
         --lf
+        --last-failed-no-failures none
       tox-tool-deps: tox
       xfail: >-
         ${{


webknjaz:
I'm not sure we were in agreement regarding what the behavior should be in the end.

webknjaz:
I self-assigned that to myself because I wanted to do some more thinking. I'm not entirely convinced if we should be addressing this at all — this patch makes the step run and result in "hey, I ran but not really, and also here's your failure — guess why". I'm worried it'd make the UX even more confusing.
I vaguely recall thinking of improving the comms, though, outputting the explanation of why and how this step may fail. And I've done that in https://github.com/tox-dev/workflow/commit/702b8ea83750bdc3839738c3be81d5c540fd4bc0 + https://github.com/tox-dev/workflow/commit/ab43dd192297b935ebca18d3d870a939d10c176e + https://github.com/tox-dev/workflow/commit/0b1e32642419392b5012b1efb66331a6babe6f79.

---

This is mainly about the UX of the CI so local testing won't demonstrate anything interesting to us. If you want to see how this changes the experience for real — you can simulate the situation within your fork (make an additional PR with the simulation against `devel` in your fork, not the upstream repo).

TheDevCodingKen:
> I self-assigned that to myself because I wanted to do some more thinking. I'm not entirely convinced if we should be addressing this at all — this patch makes the step run and result in "hey, I ran but not really, and also here's your failure — guess why". I'm worried it'd make the UX even more confusing. I vaguely recall thinking of improving the comms, though, outputting the explanation of why and how this step may fail. And I've done that in [tox-dev/workflow@702b8ea](https://github.com/tox-dev/workflow/commit/702b8ea83750bdc3839738c3be81d5c540fd4bc0) + [tox-dev/workflow@ab43dd1](https://github.com/tox-dev/workflow/commit/ab43dd192297b935ebca18d3d870a939d10c176e) + [tox-dev/workflow@0b1e326](https://github.com/tox-dev/workflow/commit/0b1e32642419392b5012b1efb66331a6babe6f79).
> 
> This is mainly about the UX of the CI so local testing won't demonstrate anything interesting to us. If you want to see how this changes the experience for real — you can simulate the situation within your fork (make an additional PR with the simulation against `devel` in your fork, not the upstream repo).

Thanks for looking into this, @webknjaz. That makes complete sense. Optimizing CI resource usage is great, but I completely agree that if it comes at the cost of DevEx and confusing CI logs, it defeats the purpose.

I really appreciate the context on the UX side of things. I'll run a simulation in my personal fork as you suggested to see exactly what the GitHub Actions output looks like with this patch applied. Let me know if you decide to go a different route with the communication output, happy to pivot or close this out based on what's best for the project's DevEx.

webknjaz:
Dunno. Perhaps, `--exitfirst` in the first run makes sense instead, provided that everything will run in the rerun.

FWIW, #21 seems mainly about the DX, not the resources, though. cc @chrismeyersfsu

TheDevCodingKen:
> Dunno. Perhaps, `--exitfirst` in the first run makes sense instead, provided that everything will run in the rerun.
> 
> FWIW, #21 seems mainly about the DX, not the resources, though. cc @chrismeyersfsu

That makes total sense regarding the DX focus. Shifting the fail-fast behavior to the initial run using --exitfirst is an interesting idea—it would definitely save the developer from waiting through a long initial run if a test breaks early.

I’ll hold off on running the simulation or pushing any alternative commits for now so you and @chrismeyersfsu can sync up on the best architectural approach for the pipeline's DevEx. Happy to implement whatever workflow you both decide is best.

webknjaz:
> That makes total sense regarding the DX focus. Shifting the fail-fast behavior to the initial run using --exitfirst is an interesting idea—it would definitely save the developer from waiting through a long initial run if a test breaks early.

But this also means that such runs may result in reporting heavily reduced coverage metrics. So I've no definitive opinion on this thus far.

TheDevCodingKen:
> But this also means that such runs may result in reporting heavily reduced coverage metrics. So I've no definitive opinion on this thus far.

Ah, that is a great point, @webknjaz. Aborting the suite early would definitely skew the coverage report since the remaining tests wouldn't execute to hit those lines of code.

It's a tricky balance between optimizing the DevEx and maintaining accurate coverage metrics! I'll leave the PR open and as-is for now while you and the team weigh the trade-offs. Absolutely no rush on my end!

webknjaz:
May I ask you what motivated you to pick up that issue?

TheDevCodingKen:
> May I ask you what motivated you to pick up that issue?

To be completely transparent, there were two main drivers, @webknjaz. 

First, I'm a dev at IBM actively seeking my next role and I wanted to keep my DevOps and infrastructure skills sharp.

Second, I had a technical interview round with the Ansible Hub team on Wed, 3/18. While I ultimately didn't land the role this time around, I wanted to get my hands dirty in a mature Ansible codebase (a) to see how the broader organization handles testing and pipeline architecture at scale, and (b) rather than just reading documentation to prepare. 

So, I was looking for an opportunity to learn, and working through these DevEx and coverage trade-offs with you has been incredibly insightful. I'm still happy to help figure out the best architectural approach for this pipeline.

PR #176, opened 2026-03-13
Add revoke Vault token functionality to OIDC lookup creds

fincamd:
AAP-64519

## Summary by CodeRabbit

* **New Features**
  * Automatic, best-effort token revocation after KV and SSH credential operations; respects Vault namespace and TLS, suppresses failures and logs a warning.

* **Bug Fixes**
  * Improved KV v2 mount/path resolution fallback when backend not specified.
  * Safer token cleanup integrated into credential flows.

* **Tests**
  * Added unit and integration tests covering revoke behavior, namespace handling, KV/SSH flows and workload-identity scenarios.

* **pre-commit notes**
  * As part of the implementation, we made vault_token a private function. This triggers a warning by pylint saying we're accessing protected members of the hashivault plugin from the test file. This is expected and will provide a solution before merging


src/awx_plugins/credentials/hashivault.py
```diff
@@ -481,6 +485,34 @@ def workload_identity_auth(**kwargs):
...
+
+        request_url = join(url, 'auth/token/revoke-self')
+
+        with CertFiles(cacert) as cert:
+            request_kwargs['verify'] = cert
+            sess.post(request_url, **request_kwargs)
+        # Best effort - don't check response status as token may already be
+        # expired/revoked, which is acceptable
+    except Exception:
+        logger.warning('Failed to revoke ephemeral Vault token')
```
fincamd: Comment for review: I'm not sure whether this should stay as warning/error/debug. It should be our best effort trying to revoke this token, so if this operation fails and no one notices, the tokens would stay alive for their TTL

src/awx_plugins/credentials/hashivault.py
```diff
@@ -481,6 +485,34 @@ def workload_identity_auth(**kwargs):
...
+            sess.headers['X-Vault-Namespace'] = kwargs['namespace']
+
+        request_url = join(url, 'auth/token/revoke-self')
+
+        with CertFiles(cacert) as cert:
+            request_kwargs['verify'] = cert
+            sess.post(request_url, **request_kwargs)
+        # Best effort - don't check response status as token may already be
+        # expired/revoked, which is acceptable
+    except Exception:
```
davemulford: Is there a pattern against using a generic `except Exception` like this? It has the potential to suppress other errors with no logging output to catch that something happened.

src/awx_plugins/credentials/hashivault.py
davemulford: Can we see if there are other cases in the project where we do this type of logging, and use what happens there?

If not let's reach out to the awx-plugin maintainers for guidance.

src/awx_plugins/credentials/hashivault.py
```diff
@@ -481,6 +485,34 @@ def workload_identity_auth(**kwargs):
     return {'role': kwargs.get('jwt_role'), 'jwt': workload_identity_token}
 
 
+def revoke_token(token: str, **kwargs):
+    """Revoke a Vault token using the token revoke-self endpoint."""
+    if not token:
+        return
+
```
davemulford, line 514: Can we add more checks to ensure we have a valid token to revoke? Things like checking that a valid CA cert was passed in `kwargs` comes to mind.

davemulford (changes requested):

melissalkelly:
Notes from review call:
1. “function with too many statements” suggestion is to add a lint ignore comment 
2. Flake8 failures should be addressed
3. Increase unit test coverage to meet 100% target, otherwise check in with awx-plugins maintainers for their feedback

src/awx_plugins/credentials/hashivault.py
webknjaz: Only ancient pieces of legacy ripped out of awx have such pattern. They should all be cleaned up. Now new code is expected/allowed to introduce this code smell.

Additionally, having more than a single instruction in try-blocks is a really bad idea that hurt maintainability as it makes it difficult to follow the control flow during inspection/troubleshooting/debugging.

> [!tip]
> #### TL;DR
> * only handle the exception that's actually known to be expected
> * only do that around one single instruction at a time (if there's a need to group multiple coupled instructions, they can be moved to a standalone function, while the *caller* would be handling the exception)
> * it's a good idea to communicate failures to the caller while shielding the calling side from the implementation details; i.e. try-except blocks may catch exceptions from underlying libraries and re-raise them as something the caller would expect as a standard interface

src/awx_plugins/credentials/hashivault.py
```diff
@@ -481,6 +485,34 @@ def workload_identity_auth(**kwargs):
     return {'role': kwargs.get('jwt_role'), 'jwt': workload_identity_token}
 
 
+def revoke_token(token: str, **kwargs):
```
webknjaz: This shouldn't accept arbitrary keword args. Make them all known. Especially, since this is a private API that's only used internally. The existing catch-all pattern, however terrible, only exists in the callables that are public API, AFAICS. I see no reason to additionally obscure the logic here.

Additionally, add a single underscore to the name to show that it's indeed private.

tests/unit/credentials/hashivault_test.py
```diff
@@ -393,3 +394,174 @@ def test_non_oidc_plugins_have_no_internal_fields(
...
+    ('should_call_post', 'token'),
+    (
+        pytest.param(False, None, id='none-token'),
+        pytest.param(False, '', id='empty-token'),
+    ),
+)
+def test_revoke_token_with_empty_token(
+    mocker: MockerFixture,
+    token: str | None,
+    *,
```
webknjaz: I don't see a reason to mark exactly one argument as kwarg-only. Do you? Technically, `pytest` will inject all of these as kwargs regardless of what you put in the signature but this just seems out-of-place and doesn't follow the commonly expected style.

tests/unit/credentials/hashivault_test.py
```diff
@@ -393,3 +394,174 @@ def test_non_oidc_plugins_have_no_internal_fields(
...
+    should_call_post: bool,
+) -> None:
+    """Test ``revoke_token()`` returns early when token is empty."""
+    mock_session = mocker.MagicMock()
+    mocker.patch('requests.Session', return_value=mock_session)
+
+    kwargs = {
```
webknjaz: No need to have a proxy variable for exactly one arg. Just pass it into the call directly.

tests/unit/credentials/hashivault_test.py
```diff
@@ -393,3 +394,174 @@ def test_non_oidc_plugins_have_no_internal_fields(
...
+        'url': 'https://vault.example.com',
+    }
+
+    hashivault.revoke_token(token, **kwargs)
+
+    if should_call_post:
+        mock_session.post.assert_called_once()
```
webknjaz: All the params in this test have `should_call_post` set to `False`. There's no need for a code branch in tests that is never hit. Additionally, it's usually not a good idea to add extra logic/branches into tests w/o a really good justification since this tends to make the tests fragile. Also note that the coverage checks will fail on dead code in tests. This shows up as missing branch coverage which is used in the respective guardrails.

tests/unit/credentials/hashivault_test.py
```diff
@@ -393,3 +394,174 @@ def test_non_oidc_plugins_have_no_internal_fields(
...
+    mocker.patch.object(
+        hashivault,
+        'CertFiles',
+        return_value=mocker.MagicMock(),
+    )
+
+    kwargs = {
+        'url': 'https://vault.example.com',
+        **extra_kwargs,
+    }
```
webknjaz, line 493: You can pass these into the call directly, no need for extra indirection since this proxy-var is only used once.

tests/unit/credentials/hashivault_test.py
```diff
@@ -393,3 +394,174 @@ def test_non_oidc_plugins_have_no_internal_fields(
...
+    )
+    mock_revoke = mocker.patch.object(hashivault, 'revoke_token')
+    mocker.patch('requests.Session')
+    mocker.patch.object(hashivault, 'CertFiles')
+    mocker.patch.object(hashivault, 'raise_for_status')
+
+    backend = getattr(hashivault, backend_func)
+
+    with contextlib.suppress(Exception):
+        backend(**backend_kwargs)
```
webknjaz: This needs justification so that the reader would know why you're doing this here.

tests/unit/credentials/hashivault_test.py
```diff
@@ -393,3 +394,174 @@ def test_non_oidc_plugins_have_no_internal_fields(
...
+@pytest.mark.parametrize(
+    ('backend_func', 'backend_kwargs'),
+    (
+        pytest.param(
+            'kv_backend',
+            {
+                'url': 'https://vault.example.com',
+                'workload_identity_token': 'jwt_token',
+                'jwt_role': 'test_role',
+                'default_auth_path': 'jwt',
```
webknjaz: The immutable portion of the args could move into the test body, this would reduce duplication.

tests/unit/credentials/hashivault_test.py
```diff
@@ -393,3 +394,174 @@ def test_non_oidc_plugins_have_no_internal_fields(
...
+    """Test backend functions revoke token in finally block for OIDC auth."""
+    mock_handle_auth = mocker.patch.object(
+        hashivault,
+        'handle_auth',
+        return_value='test_token',
+    )
+    mock_revoke = mocker.patch.object(hashivault, 'revoke_token')
+    mocker.patch('requests.Session')
+    mocker.patch.object(hashivault, 'CertFiles')
+    mocker.patch.object(hashivault, 'raise_for_status')
```
webknjaz: Remember to always use autospec in your mocks.

tests/unit/credentials/hashivault_test.py
```diff
@@ -393,3 +394,174 @@ def test_non_oidc_plugins_have_no_internal_fields(
...
+        assert mock_session.headers[header] == value
+    mock_post.assert_called_once()
+    assert 'auth/token/revoke-self' in mock_post.call_args[0][0]
+
+
+def test_revoke_token_handles_exceptions(mocker: MockerFixture) -> None:
+    """Test ``revoke_token()`` handles exceptions gracefully."""
+    mock_session = mocker.MagicMock()
+    mock_session.headers = {}
+    mock_session.post.side_effect = Exception('Network error')
```
webknjaz: Best simulate actual exception that can really happen, not arbitrary base exceptions.

tests/unit/credentials/hashivault_test.py
```diff
@@ -393,3 +394,174 @@ def test_non_oidc_plugins_have_no_internal_fields(
         field for field in plugin.inputs['fields'] if field.get('internal')
     ]
     assert internal_fields == []
+
+
+@pytest.mark.parametrize(
+    ('should_call_post', 'token'),
```
webknjaz: Unless you intend to add more params with `should_call_post` set to `True`, it's redundant.

tests/unit/credentials/hashivault_test.py
```diff
@@ -393,3 +394,174 @@ def test_non_oidc_plugins_have_no_internal_fields(
...
+    kwargs = {
+        'url': 'https://vault.example.com',
+    }
+
+    # Should not raise exception
+    hashivault.revoke_token('test_token', **kwargs)
+
+    mock_logger.warning.assert_called_once_with(
+        'Failed to revoke ephemeral Vault token',
+    )
```
webknjaz: This is not the right way of checking for warnings. Instead, use `pytest`'s built-in facility for exactly this — the `caplog` fixture that's provided out-of-the box.

tests/unit/credentials/hashivault_test.py
```diff
@@ -393,3 +394,174 @@ def test_non_oidc_plugins_have_no_internal_fields(
...
+        hashivault,
+        'handle_auth',
+        return_value='test_token',
+    )
+    mock_revoke = mocker.patch.object(hashivault, 'revoke_token')
+    mocker.patch('requests.Session')
+    mocker.patch.object(hashivault, 'CertFiles')
+    mocker.patch.object(hashivault, 'raise_for_status')
+
+    backend = getattr(hashivault, backend_func)
```
webknjaz: Instead of using dynamic trampoline logic for looking up things that are always there and don't need fallbacks, pass these objects as params by accessing the attributes directly, like `test_non_oidc_plugins_have_no_internal_fields` does, for example.

src/awx_plugins/credentials/hashivault.py
```diff
@@ -16,6 +18,8 @@
 from .plugin import CertFiles, CredentialPlugin, raise_for_status
 
 
+logger = getLogger(__name__)
```
webknjaz: Try to avoid creating public names in the global namespace. Plus, it's really uncommon to import `getLogger` separately.
```suggestion
_logger = _logging.getLogger(__name__)
```

src/awx_plugins/credentials/hashivault.py
```diff
@@ -481,6 +485,34 @@ def workload_identity_auth(**kwargs):
     return {'role': kwargs.get('jwt_role'), 'jwt': workload_identity_token}
 
 
+def revoke_token(token: str, **kwargs):
+    """Revoke a Vault token using the token revoke-self endpoint."""
+    if not token:
+        return
+
+    try:
+        url = join(kwargs['url'], 'v1')
```
webknjaz: String path helpers mustn't be used to handle URLs. Use `urllib.parse.urljoin` instead.

src/awx_plugins/credentials/hashivault.py
```diff
@@ -4,6 +4,8 @@
 import os
 import pathlib
 import time
+from logging import getLogger
+from os.path import join
```
webknjaz: There's `urllib.parse.urljoin` in stdlib for the purpose this is currently abused for. You don't need this import.
```suggestion
```

src/awx_plugins/credentials/hashivault.py
```diff
@@ -4,6 +4,8 @@
 import os
 import pathlib
 import time
+from logging import getLogger
```
webknjaz: Let's avoid adding more public things into the global namespace.
```suggestion
import logging as _logging
```

src/awx_plugins/credentials/hashivault.py
```diff
@@ -481,6 +485,34 @@ def workload_identity_auth(**kwargs):
...
+    try:
+        url = join(kwargs['url'], 'v1')
+        cacert = kwargs.get('cacert')
+
+        request_kwargs = {'timeout': 10}
+
+        sess = requests.Session()
+        sess.mount(url, requests.adapters.HTTPAdapter(max_retries=3))
+        sess.headers['X-Vault-Token'] = token
+        if kwargs.get('namespace'):
```
webknjaz: You probably meant "if it's missing" instead of "if it's falsy", right? An empty string ain't the same as `None`.

src/awx_plugins/credentials/hashivault.py
webknjaz: Here's a better signature that makes it obvious what's to be passed, makes sure it's always called w/ the kwargs syntax and helps understand typing expectations:
```suggestion
def _revoke_token(*, token: str, url: str, cacert: str | None, namespace: str) -> None:
```

src/awx_plugins/credentials/hashivault.py
```diff
@@ -522,127 +554,145 @@ def method_auth(**kwargs):
...
-        sess.headers['X-Vault-Namespace'] = kwargs['namespace']
-
-    if api_version == 'v2':
-        if kwargs.get('secret_version'):
-            request_kwargs['params'] = {  # type: ignore[assignment]  # FIXME
-                'version': kwargs['secret_version'],
-            }
-        if secret_backend:
-            path_segments = [secret_backend, 'data', secret_path]
+def kv_backend(**kwargs):  # noqa: PLR0915
```
webknjaz: Unjustified linting suppressions must not be added. They must always be paired with explanatory comments, expecially those without human-readable identifiers with a bunch of numbers that don't really mean anything to the reader.

---

Anyway, this one doesn't look justified regardless. It seems like you want to have a cleanup in a bunch of high-level functions. This can likely be solved elegantly via `@contextlib.contextmanager`, which produces not only as a CM but the result can also be used as a function decorator too, saving us a lot of unnecessary indentation.

While writing this I'd drafted smth like this:
```python
import contextlib as _ctx
import functools as _ft

# the above goes to the top of the module ^
...


@_ctx.contextmanager
def _clean_up_oidc_token(*args, **kwargs):
    try:
        yield
    finally:
        if 'workload_identity_token' not in kwargs:
            return

        _revoke_token(
            token=...,
            url=kwargs['url'],
            cacert=kwargs.get('cacert'),
            namespace=kwargs.get('namespace'),
        )


def _auto_revoke_oidc_token(call_the_decorated_function, /):
    @_ft.wraps(call_the_decorated_function)
    def decorate_the_function_with_cleanup(*args, **kwargs):
        with _clean_up_oidc_token(*args, **kwargs):
            call_the_decorated_function(*args, **kwargs)

    return decorate_the_function_with_cleanup


@_auto_revoke_oidc_token  # <-- apply this to other callables the same way
def kv_backend(**kwargs):
    ...  # the original contents here
```

But then I realized that the token's only available within the plugin functions. So next I realized that the cleanup is actually quite coupled with the auth handling logic so `handle_auth()` should go into a CM:
```python
import contextlib as _ctx
import functools as _ft

# the above goes to the top of the module ^
...


@_ctx.contextmanager
def _lifetime_bound_token(**kwargs) -> str:
    token_is_revocable = 'workload_identity_token' in kwargs  # e.g. ephemeral vault token

    auth_token = handle_auth(**kwargs)

    try:
        yield auth_token
    finally:
        if not token_is_revocable:
            return

        _revoke_token(
            token=...,
            url=kwargs['url'],
            cacert=kwargs.get('cacert'),
            namespace=kwargs.get('namespace'),
        )


def _inject_auth_token(call_the_decorated_function, /):
    @_ft.wraps(call_the_decorated_function)
    def decorate_the_function_with_cleanup(**kwargs):
        with _lifetime_bound_token(**kwargs) as http_auth_token:
            call_the_decorated_function(token=http_auth_token, **kwargs)

    return decorate_the_function_with_cleanup


@_inject_auth_token
def kv_backend(*, token: str, **kwargs):
    ...  # the original contents here, with the `handle_auth()` call removed


@_inject_auth_token
def ssh_backend(*, token: str, **kwargs):
    ...  # the original contents here, with the `handle_auth()` call removed
```

> [!important]
> The above is sketched in-browser so it may contain typos/cosmetic linting errors
> but should be good to go otherwise.

src/awx_plugins/credentials/hashivault.py
webknjaz: If this is determined to need some kind of a base exception (like what this patch does now), we should re-raise it post logging. Only specific known exception may have a justifiable cause for full suppression.

webknjaz: This would probably be re-raised as a `RuntimeError` or something like that.

src/awx_plugins/credentials/hashivault.py
```diff
@@ -522,127 +554,145 @@ def method_auth(**kwargs):
...
-
-    json = response.json()
-    if api_version == 'v2':
-        json = json['data']
-
-    if secret_key:
-        try:
-            if (
-                (secret_key != 'data')
-                and (  # noqa: S105; not a password
```
webknjaz, line 591: Do not remove any justification comments as this causes suppressions to loose context, making them seem unjustified in the future. It's really bad for maintainability.

webknjaz:
Since it's WIP, I've marked it as a draft to make it visible that it's incomplete. It's a good idea to do so in general and undraft when requesting reviews.

src/awx_plugins/credentials/hashivault.py
fincamd: i will then leave it with no try-except block and let any HTTP request failure to float to the caller. THanks all!

webknjaz: Well, it's known which exceptions the `requests` library produces — those could be captured and turned into `RuntimeError` or something more specific in some cases.

src/awx_plugins/credentials/hashivault.py
fincamd: Much a cleaner solution, thanks for taking the time to get this review in @webknjaz . I appreciate it a lot!!

src/awx_plugins/credentials/hashivault.py
```diff
@@ -481,6 +483,58 @@ def workload_identity_auth(**kwargs):
...
+            )
+
+
+def _inject_auth_token_with_revokation(decorated_function, /):
+    @_functools.wraps(decorated_function)
+    def _decorate_the_function_with_revokation(**kwargs):
+        with _handle_vault_token_revokation(**kwargs) as http_auth_token:
+            decorated_function(token=http_auth_token, **kwargs)
+
+    return _decorate_the_function_with_revokation
```
webknjaz: Google tells me it's k->c, grammar-wise: https://www.merriam-webster.com/dictionary/revocation.

Also, there's no need to use underscored vars within the `_inject_auth_token_with_revocation()` function since they are scoped to that and won't leak externally.

src/awx_plugins/credentials/hashivault.py
```diff
@@ -522,8 +576,9 @@ def method_auth(**kwargs):
     return token
 
 
+@_inject_auth_token_with_revokation
 def kv_backend(**kwargs):
```
webknjaz: Why didn't you go for an explicit `token` arg? It's pretty much guaranteed to be passed since it's an implementation detail.

src/awx_plugins/credentials/hashivault.py
```diff
@@ -566,7 +621,9 @@ def kv_backend(**kwargs):
     else:
         path_segments = [secret_path]
 
-    request_url = urljoin(url, '/'.join(['v1'] + path_segments)).rstrip('/')
+    request_url = urljoin(url, '/'.join(['v1'] + path_segments)).rstrip(
+        '/',
+    )
```
webknjaz: Could you revert unrelated formatting changes?

src/awx_plugins/credentials/hashivault.py
```diff
@@ -596,12 +653,15 @@ def kv_backend(**kwargs):
                 return json['data']['data'][secret_key]
             return json['data'][secret_key]
         except KeyError:
-            raise RuntimeError(f'{secret_key} is not present at {secret_path}')
+            raise RuntimeError(
+                f'{secret_key} is not present at {secret_path}',
+            )
```
webknjaz, line 690: Reverting this will also make the patch smaller and easier to review.

webknjaz:
> [!important]
> The flake8 exception under Python 3.11.15 in one of the jobs ain't your fault and I've fixed it on `devel` via #177. But the 3.14 job running the same shows legit violations you'll need to fix.

---

There's also some renaming needed in the test module, obviously + check the remaining suggestions.

webknjaz:
Looks like your most recent force-push lost some of the good changes that existed in this branch a few days ago: https://github.com/ansible/awx-plugins/compare/2e90e464e1f282c23727284e644805b1d2ecb9a4..275b035e40ce7bd9dca32c1d7d1916d9fbb4066d

src/awx_plugins/credentials/hashivault.py
```diff
@@ -481,6 +484,34 @@ def workload_identity_auth(**kwargs):
     return {'role': kwargs.get('jwt_role'), 'jwt': workload_identity_token}
 
 
+def revoke_token(token: str | None, **kwargs):
```
webknjaz: It's probably a bad idea to accept `None` here. I don't see why it'd be legitimate to call this helper if there's no token to revoke.

src/awx_plugins/credentials/hashivault.py
```diff
@@ -4,6 +4,7 @@
 import os
 import pathlib
 import time
+from logging import getLogger
```
dleehr: Can you help me understand why we need logging in this plugin? This addition stands out because none of the other credential plugins do any logging and `hashivault.py` didn't do any logging prior to this.

src/awx_plugins/credentials/hashivault.py
dleehr: Yeah I was thinking of a similar contextmanager function:


```python
import contextlib as _ctx

@_ctx.contextmanager
def vault_token(**kwargs):
    """Context manager that yields a Vault token and revokes it on exit if obtained via workload identity."""
    token = handle_auth(**kwargs)
    try:
        yield token
    finally:
        # Only revoke tokens obtained via workload identity authentication
        if kwargs.get('workload_identity_token'):
            url = urljoin(kwargs['url'], 'v1/auth/token/revoke-self')
            sess = requests.Session()
            sess.headers['X-Vault-Token'] = token
            if kwargs.get('namespace'):
                sess.headers['X-Vault-Namespace'] = kwargs['namespace']
            with CertFiles(kwargs.get('cacert')) as cert:
                resp = sess.post(url, verify=cert, timeout=30)
            resp.raise_for_status() # see my other comment
```

Then 

```python
token = handle_auth(**kwargs)
...
```

becomes

```python
with vault_token(**kwargs) as token:
    ...
```

src/awx_plugins/credentials/hashivault.py
dleehr: Not sure what the resolution was here around exception handling, but I have some feedback.

I do not think we should suppress this exception **or** log anything. I think we should fail (using `raise_for_status`) like we do for every other `requests` call. We're implementing revocation and setting the expectation that vault tokens are revoked, so if we suppress failures, we're undermining that expectation.

In the scenario where tokens are **not** revoked, the user will only know that if they review the logs. As a security-minded user I would be surprised / concerned to discover that tokens weren't being revoked.

The proposal states that this must raise an exception:

> Per P2R3, the following error scenarios must be handled via raising exceptions in the plugin code.
> - ...
> - Token Revocation Failure: Errors when revoking Vault client token

Is there a reason the A/C in the story is different?

If we need to provide some flexibility here, I have a suggestion.

dleehr (commented):
Had some feedback on implementation around using a context manager (but not as elegant as @webknjaz), but moreso a question about suppressing all exceptions on revoke.

I left a line-level comment there, but not sure if that's already being considered.

src/awx_plugins/credentials/hashivault.py
fincamd: I just added this logging as part of the effort of not silencing any exception that might arise. From what I saw  during testing, the user would get the exception trace and that is not appropriate, so I figured it'd be a good idea to log this as a warning/error

fincamd: I have removed this as part of this commit https://github.com/ansible/awx-plugins/pull/176/changes/24c5a68b6a7643eabf7212525d428fa8b39e1de4

src/awx_plugins/credentials/hashivault.py
fincamd: True, I was so focused trying to fix pre-commit issues that I didn't double check this. Thanks! Now that the functionality is implemented as a context_manager I have removed this bit

fincamd:
>
Looks like your most recent force-push lost some of the good changes that existed in this branch a few days ago: https://github.com/ansible/awx-plugins/compare/2e90e464e1f282c23727284e644805b1d2ecb9a4..275b035e40ce7bd9dca32c1d7d1916d9fbb4066d

Hi @webknjaz, thanks for all of your detailed feedback! I want to push this to merge as is, I force-pushed the branch to an earlier commit with just the necessary changes plus latest reviews. The functionality works and I only need approvals from code owners/stakeholders. I will follow-up with more changes refining this with your advice in another PR

src/awx_plugins/credentials/hashivault.py
```diff
@@ -522,127 +543,131 @@ def method_auth(**kwargs):
...
-        sess.headers['X-Vault-Namespace'] = kwargs['namespace']
-
-    if api_version == 'v2':
-        if kwargs.get('secret_version'):
-            request_kwargs['params'] = {  # type: ignore[assignment]  # FIXME
-                'version': kwargs['secret_version'],
-            }
-        if secret_backend:
-            path_segments = [secret_backend, 'data', secret_path]
+def kv_backend(**kwargs):  # noqa: PLR0915
```
dleehr: @fincamd I see you resolved the previous comment but I don't think @webknjaz's feedback has been addressed:

> Unjustified linting suppressions must not be added

Since you adopted a context manager, this function shouldn't trigger "too many statements" anymore. Please remove the `# noqa: PLR0915`. It should not be needed.

src/awx_plugins/credentials/hashivault.py
```diff
@@ -522,127 +543,131 @@ def method_auth(**kwargs):
...
+                # https://www.vaultproject.io/api/secret/kv/kv-v2.html#read-secret-version
+                path_segments = [mount_point, 'data'] + path
+        elif secret_backend:
+            path_segments = [secret_backend, secret_path]
         else:
-            try:
-                mount_point, *path = pathlib.Path(
-                    secret_path.lstrip(os.sep),
-                ).parts
-                '/'.join(path)
```
dleehr, line 559: It's a bit hard to tell without ignoring whitespace (https://github.com/ansible/awx-plugins/pull/176/changes?w=1), but this looks like a good catch - `'/'.join(path)` was just dead code so just that line is deleted:

https://github.com/ansible/awx-plugins/blob/c464cf276b9344056d7f47090bf361c2a8d85b45/src/awx_plugins/credentials/hashivault.py#L555-L560

src/awx_plugins/credentials/hashivault.py
dleehr: Yeah looks like indentation and reformatting deleted this important comment, please restore.

src/awx_plugins/credentials/hashivault.py
```diff
@@ -481,6 +483,25 @@ def workload_identity_auth(**kwargs):
     return {'role': kwargs.get('jwt_role'), 'jwt': workload_identity_token}
 
 
+@_ctx.contextmanager
+def vault_token(**kwargs) -> Generator[str, None, None]:
```
dleehr: This should fix `error: Type of decorated function contains type "Any"`.

```suggestion
def vault_token(**kwargs: str) -> Generator[str, None, None]:
```

For `**kwargs`, the type annotation [provides the type for the values](https://adamj.eu/tech/2021/05/11/python-type-hints-args-and-kwargs/), so `str` here means we treat `kwargs` like a `dict[str,str]`.

I looked at the values we pull out of `kwargs` here (and in `handle_auth`) and they're all strings, so I think this is a reasonable annotation.

tests/unit/credentials/hashivault_test.py
```diff
@@ -393,3 +394,250 @@ def test_non_oidc_plugins_have_no_internal_fields(
...
+        'default_auth_path': 'jwt',
+        **extra_kwargs,
+    }
+
+    with hashivault.vault_token(**kwargs) as token:
+        assert token == 'test_token'
+
+    mock_handle_auth.assert_called_once_with(**kwargs)
+    for header, contents in expected_headers.items():
+        assert mock_session.headers[header] == contents
```
dleehr:
```suggestion
    for header, header_value in expected_headers.items():
        assert mock_session.headers[header] == header_value
```

src/awx_plugins/credentials/hashivault.py
```diff
@@ -522,127 +543,131 @@ def method_auth(**kwargs):
...
+        sess.mount(kwargs['url'], requests.adapters.HTTPAdapter(max_retries=5))
+        sess.headers['Authorization'] = f'Bearer {token}'
+        # Compatibility header for older installs of Hashicorp Vault
+        sess.headers['X-Vault-Token'] = token
+        if kwargs.get('namespace'):
+            sess.headers['X-Vault-Namespace'] = kwargs['namespace']
+
+        if api_version == 'v2':
+            if kwargs.get('secret_version'):
+                request_kwargs['params'] = {  # type: ignore[assignment]  # FIXME
```
dleehr: Suggesting to delete the FIXME since it makes the line 81 characters now and fails the linter. 

```suggestion
                request_kwargs['params'] = {  # type: ignore[assignment]
```

dleehr (approved):
LGTM, approving with a suggestion to fix that `error: Type of decorated function contains type "Any"` but please address the feedback from @webknjaz to remove unneeded linting suppressions, restore comments, and provide comments around unexpected logic.

tests/unit/credentials/hashivault_test.py
dleehr: Since the linter complained about `contents` as a variable name

src/awx_plugins/credentials/hashivault.py
dleehr: Also, I didn't have it in my suggested example but we should name the function `_vault_token` since it's intended to be private/internal.

fincamd: I tried that too yesterday addressing @webknjaz's feedback. However, it will trigger a complaint from the linter saying that we shouldn't access a protected member from the test cases. So we have to again, choose an approach to make the linter happy

src/awx_plugins/credentials/hashivault.py
fincamd: Nice, this fixes that issue. Thanks Dan

src/awx_plugins/credentials/hashivault.py
fincamd: Yes! That was dead code, I deleted it in this branch

tests/unit/credentials/hashivault_test.py
fincamd: @webknjaz How would you type the function parameter? I've tried these options and neither of them satisfy the linter:

```
-     backend_func: object,
-     backend_func: _t.Callable[..., str],
-     backend_func: _t.Callable[..., _t.Any],
-     backend_func: _t.Callable[dict[str, str], str],
```

dleehr (approved):

src/awx_plugins/credentials/hashivault.py
webknjaz: @fincamd the violation within the project is fine — we can have a justified suppression in the tests while still communicating that the thing is private to the downstream consumers. A case could be made for disabling that rule for the test folder, for example. Just make sure to document such cases in an issue or a top-level comment in this PR so I would be able to track and address them independently.

webknjaz:
@fincamd could you rebase instead of adding merge commits into the branch?

src/awx_plugins/credentials/hashivault.py
webknjaz: It's usually a good idea to have an explanation for each suppression. And that would mean putting a bigger comment on a separate line, which can start with a FIXME.

tests/unit/credentials/hashivault_test.py
webknjaz: @dleehr it'd be useful to post context of the motivation. "a linter complained" doesn't really explain what rule you violated, or which linter hinted you to fix this. When reading through the old comment threads, it's interesting to understand the root causes.

davemulford (approved):

src/awx_plugins/credentials/hashivault.py
```diff
@@ -1,9 +1,11 @@
 # FIXME: the following violations must be addressed gradually and unignored
 # mypy: disable-error-code="arg-type, no-untyped-call, no-untyped-def"
 
+import contextlib as _ctx
 import os
 import pathlib
 import time
+from collections.abc import Generator
```
webknjaz: Use

```suggestion
import collections.abc as _c
```
instead as a convention similar to importing typing. This will make sure that no new public objects are exposed.

src/awx_plugins/credentials/hashivault.py
```diff
@@ -481,6 +483,25 @@ def workload_identity_auth(**kwargs):
     return {'role': kwargs.get('jwt_role'), 'jwt': workload_identity_token}
 
 
+@_ctx.contextmanager
+def vault_token(**kwargs: str) -> Generator[str, None, None]:
```
webknjaz: It's much easier to decorate such generators w/ the `_c.Iterator[str]` return type.

src/awx_plugins/credentials/hashivault.py
```diff
@@ -481,6 +483,25 @@ def workload_identity_auth(**kwargs):
...
+    """Context manager that yields a Vault token and revokes it on exit if obtained via workload identity."""
+    token = handle_auth(**kwargs)
+    try:
+        yield token
+    finally:
+        # Only revoke tokens obtained via workload identity authentication
+        if kwargs.get('workload_identity_token'):
```
webknjaz: Instead of having this much indentation, use the `if not X: return` guard check, which allows dedenting 8 lines, making it easier on the reader.

Also, why do you use `dict.get()` for this? Is there a difference between `None` and `''`? Does mere presence of the arg mean that the context is OIDC if so, using a key presence check would be better semantically.

Additionally, it's better to name whatever check is here by doing it at the beginning of the function and assigning a readable value. If you name it well, there would be no need for a code comment as the name would communicate the semantics of the check instead. `if not is_oidc_context: return` would read better than "if some key's value is falsy or it's absent, do this bunch of logic below, and oh by the way, check that comment above to understand the semantics".

src/awx_plugins/credentials/hashivault.py
webknjaz: Also, why is this API still public?

webknjaz: Can we list the args that are used and guaranteed in the signature, please? Or at least, extract them into vars w/ a regular dict lookup instead of the variant that falls back to `None`?

src/awx_plugins/credentials/hashivault.py
webknjaz: Let's not make unrelated functional or formatting changes in this branch.

src/awx_plugins/credentials/hashivault.py
```diff
@@ -481,6 +483,25 @@ def workload_identity_auth(**kwargs):
...
+            url = urljoin(kwargs['url'], 'v1/auth/token/revoke-self')
+            sess = requests.Session()
+            sess.headers['X-Vault-Token'] = token
+            if kwargs.get('namespace'):
+                sess.headers['X-Vault-Namespace'] = kwargs['namespace']
+            with CertFiles(kwargs.get('cacert')) as cert:
+                resp = sess.post(url, verify=cert, timeout=30)
+            resp.raise_for_status()
```
webknjaz: This could probably be in a separate helper function since this piece of logic looks coupled.

src/awx_plugins/credentials/hashivault.py
```diff
@@ -523,126 +544,130 @@ def method_auth(**kwargs):
...
-        sess.headers['X-Vault-Namespace'] = kwargs['namespace']
-
-    if api_version == 'v2':
-        if kwargs.get('secret_version'):
-            request_kwargs['params'] = {  # type: ignore[assignment]  # FIXME
-                'version': kwargs['secret_version'],
-            }
-        if secret_backend:
-            path_segments = [secret_backend, 'data', secret_path]
+    with vault_token(**kwargs) as token:
```
webknjaz: I'd rather not keep a function-wide CM so that it doesn't cause unreasonable levels of indentation. Also, the CM itself isn't something that fits well in the function semantics and I believe a decorator is the best place for it.

tests/unit/credentials/hashivault_test.py
```diff
@@ -395,6 +396,256 @@ def test_non_oidc_plugins_have_no_internal_fields(
...
+    """Test ``vault_token`` context manager doesn't revoke token without workload_identity_token."""
+    mock_handle_auth = mocker.patch.object(
+        hashivault,
+        'handle_auth',
+        return_value='test_token',
+    )
+    mock_session = mocker.MagicMock()
+    mocker.patch('requests.Session', return_value=mock_session)
+
+    kwargs = {
```
webknjaz: Let's use a less generic name for this var.

tests/unit/credentials/hashivault_test.py
```diff
@@ -395,6 +396,256 @@ def test_non_oidc_plugins_have_no_internal_fields(
...
+def test_vault_token_no_workload_identity(
+    mocker: MockerFixture,
+) -> None:
+    """Test ``vault_token`` context manager doesn't revoke token without workload_identity_token."""
+    mock_handle_auth = mocker.patch.object(
+        hashivault,
+        'handle_auth',
+        return_value='test_token',
+    )
+    mock_session = mocker.MagicMock()
```
webknjaz: Why is this mock magic?

tests/unit/credentials/hashivault_test.py
```diff
@@ -395,6 +396,256 @@ def test_non_oidc_plugins_have_no_internal_fields(
...
+    mocker.patch('requests.Session', return_value=mock_session)
```
webknjaz: Does this even need a `return_value`?

tests/unit/credentials/hashivault_test.py
```diff
@@ -395,6 +396,256 @@ def test_non_oidc_plugins_have_no_internal_fields(
     assert internal_fields == []
 
 
+def test_vault_token_no_workload_identity(
+    mocker: MockerFixture,
+) -> None:
+    """Test ``vault_token`` context manager doesn't revoke token without workload_identity_token."""
+    mock_handle_auth = mocker.patch.object(
```
webknjaz: Remember to set the spec whenever creating mocks. It can be an autospec most of the time.

https://nedbatchelder.com/blog/202202/why_your_mock_still_doesnt_work

tests/unit/credentials/hashivault_test.py
```diff
@@ -395,6 +396,256 @@ def test_non_oidc_plugins_have_no_internal_fields(
...
+    """Test ``vault_token`` context manager raises when token revocation fails."""
+    mock_handle_auth = mocker.patch.object(
+        hashivault,
+        'handle_auth',
+        return_value='test_token',
+    )
+    mock_session = mocker.MagicMock()
+    mock_session.headers = {}
+    mock_response = mocker.MagicMock()
+    mock_response.raise_for_status.side_effect = Exception('Revocation failed')
```
webknjaz: Don't use exceptions that never happen IRL. Use something that's known to be raised by the library being called.

tests/unit/credentials/hashivault_test.py
```diff
@@ -395,6 +396,256 @@ def test_non_oidc_plugins_have_no_internal_fields(
...
+        'workload_identity_token': 'jwt_token',
+        'jwt_role': 'test_role',
+        'default_auth_path': 'jwt',
+    }
+
+    def _use_vault_token() -> None:
+        with hashivault.vault_token(**kwargs) as token:
+            assert token == 'test_token'
+
+    with pytest.raises(Exception, match='Revocation failed'):
```
webknjaz: Asserting for base exceptions is not very useful. Best be specific.

tests/unit/credentials/hashivault_test.py
webknjaz: The last one should be fine. But it won't work unless you actually type the definition itself.

src/awx_plugins/credentials/hashivault.py
fincamd: Because the tests that use the function calling it from the test script trigger pre-commit ci warnings saying we access a protected member of the hashivault.py module. Any ideas to solve this while keeping the function private?

src/awx_plugins/credentials/hashivault.py
webknjaz: @dleehr my suggestion has abstraction layers separated and doesn't put things from multiple layers together. It's verbose to make said layers obvious. And the external decorator is a better fit for the handlers because (1) it doesn't add indentation which is better for maintainability, not just for the diff in review and (2) the auth is a common separate process so it doesn't make sense semantically to make it a part of the handlers but something that is performed externally. This can also be scaled to other plugins by moving to a helper module later if needed. Note that explicit args is better then the unclear semantics of `dict.get('thing_name', None)` which is one of "arg is missing, or arg is None, or arg is an empty string, or arg is anything falsy" that's not really explained in the condensed style that is overly indented internally as well.

I mentioned this here https://github.com/ansible/awx-plugins/pull/176/changes#r2988429724 and here https://github.com/ansible/awx-plugins/pull/176/changes#r2988429724.

src/awx_plugins/credentials/hashivault.py
webknjaz: @dleehr I don't have an opinion on logging (it's probably a good idea to have it for troubleshooting purposes, *as long as the exception bubbles up to the outer layers*) but in general, I'd expect there to be some sort of a common exception framework that the plugins would use eventually. And it'd be the `awx`s responsibility to process them and log/show the unexpected ones. There's also a case to be made that a good place for the unexpected error normalization layer is the `awx_plugins.interfaces` library which would provide a shared exception hierarchy to be used by both `awx and `awx-plugins-core`.

webknjaz:
@fincamd I checked the red PR checks and have some explanation for those.

That unrelated indentation makes a whole bunch of lines a part of the patch region that Codecov+Coveralls will expect to be covered. This is likely what hurts the measured patch coverage metric. It's best to avoid making changes to parts of the previously uncovered code base (like moving/indenting it) unless you also include tests to ensure it's fully covered.

So I don't see this approach as justified.

src/awx_plugins/credentials/hashivault.py
webknjaz: Yeah, I think I commented on some other thread that it's okay to have a local suppression but if you document such cases as a top level PR comment, it'll be easier for me to address this systematically. This specific rule seems to be a good candidate for disabling in tests (and keeping enabled in the runtime code).

src/awx_plugins/credentials/hashivault.py
```diff
@@ -523,126 +544,130 @@ def method_auth(**kwargs):
...
-    sess.mount(url, requests.adapters.HTTPAdapter(max_retries=5))
-    sess.headers['Authorization'] = f'Bearer {token}'
-    # Compatibility header for older installs of Hashicorp Vault
-    sess.headers['X-Vault-Token'] = token
-    if kwargs.get('namespace'):
-        sess.headers['X-Vault-Namespace'] = kwargs['namespace']
-
-    if api_version == 'v2':
-        if kwargs.get('secret_version'):
-            request_kwargs['params'] = {  # type: ignore[assignment]  # FIXME
```
webknjaz, line 549: The lost comment needs to be re-explained, obviously. However, I advise against indenting entire function bodies.

src/awx_plugins/credentials/hashivault.py
webknjaz: It may be a good idea to record this as an issue to research and possibly implement globally.

src/awx_plugins/credentials/hashivault.py
fincamd: Great, I'll make the function private and make the comment somewhere in the PR

src/awx_plugins/credentials/hashivault.py
fincamd: >Instead of having this much indentation, use the if not X: return guard check, which allows dedenting 8 lines, making it easier on the reader.

This will trigger this error:
```
B012 `return` inside `finally` blocks cause exceptions to be silenced
   --> src/awx_plugins/credentials/hashivault.py:495:13
    |
493 |         # Only revoke tokens obtained via workload identity authentication
494 |         if 'workload_identity_token' not in kwargs:
495 |             return
    |             ^^^^^^
``` 

I have addressed the readability issues following your advice

src/awx_plugins/credentials/hashivault.py
webknjaz: Could you also return to the previous approach with clear abstractions, while on it?

src/awx_plugins/credentials/hashivault.py
webknjaz: Ah, I forgot about that quirk. Makes sense. So I'd still have the check result in a named var but keep only one instruction in the conditional block — this makes sure there's not many indented things but also keeps the abstraction layers separate: the token retrieval logic is scoped in a helper function and there's no reason for the token revocation (semantically similar but opposite) to leak its implementation details into the top-level calling code. I should've made my thinking more verbose earlier to justify the idea..

src/awx_plugins/credentials/hashivault.py
fincamd: This has been addressed in this commit 410a0a7f834305308b5ee96375583a01e686d147

src/awx_plugins/credentials/hashivault.py
fincamd: Context manager implementation, unindenting of the functions and explicit parameter listing in 410a0a7f834305308b5ee96375583a01e686d147

src/awx_plugins/credentials/hashivault.py
fincamd: Done in 410a0a7f834305308b5ee96375583a01e686d147

src/awx_plugins/credentials/hashivault.py
fincamd: You mean enabling logging in awx-plugins so that it filters through to awx logging as well? @webknjaz

src/awx_plugins/credentials/hashivault.py
fincamd: Nice, this last commit that fixes mypy errors in the decorator function (b69f321b8e26ab79d3ee9653391ab1071e5c3172) will trigger this error of accessing private/protected members from the tests. We can decide to suppress that, I'll defer to you

src/awx_plugins/credentials/hashivault.py
fincamd: Well, it's just a line deletion... I think we can go forward. It was just dead code anyways

src/awx_plugins/credentials/hashivault.py
fincamd: Done in 410a0a7f8

src/awx_plugins/credentials/hashivault.py
fincamd: We are not following this approach and using ParamSpec instead (b69f321b8e26ab79d3ee9653391ab1071e5c3172)

src/awx_plugins/credentials/hashivault.py
fincamd: Made these changes in 24c5a68b6

src/awx_plugins/credentials/hashivault.py
fincamd: Function parameters made explicit in 410a0a7f834305308b5ee96375583a01e686d147.
I will make a note in the PR description saying there will be pre-commit errors due to this

fincamd: Made the list of arguments explicit here 410a0a7f834305308b5ee96375583a01e686d147. It triggers some mypy errors of "too many arguments"

src/awx_plugins/credentials/hashivault.py
fincamd: Done in 410a0a7f8

src/awx_plugins/credentials/hashivault.py
fincamd: Done in 410a0a7f834305308b5ee96375583a01e686d147

tests/unit/credentials/hashivault_test.py
fincamd: Right, I figured how to remove this and fix the tests in 5126a1f3bab53ccb15fd09ee43e23d85c979f79d

tests/unit/credentials/hashivault_test.py
fincamd: Done in ab17d6c23b7b3faf051cfc5c8d5fed00cef016d2

tests/unit/credentials/hashivault_test.py
fincamd: Done here f60a6b5bc3b76e26ef591321fc2f9a4f775cfda2

tests/unit/credentials/hashivault_test.py
fincamd: Done f07295ecb580ce67eeb162d046c043ee66faa4b7

tests/unit/credentials/hashivault_test.py
fincamd: Removed as part of ab17d6c23b7b3faf051cfc5c8d5fed00cef016d2

tests/unit/credentials/hashivault_test.py
fincamd: Removed as part of https://github.com/ansible/awx-plugins/commit/ab17d6c23b7b3faf051cfc5c8d5fed00cef016d2

tests/unit/credentials/hashivault_test.py
fincamd: Removed as part of https://github.com/ansible/awx-plugins/commit/ab17d6c23b7b3faf051cfc5c8d5fed00cef016d2

tests/unit/credentials/hashivault_test.py
fincamd: Done in 3a4b7991

tests/unit/credentials/hashivault_test.py
fincamd: I'm catching the Exception that's being used in the handle_auth function in hashivault.py https://github.com/ansible/awx-plugins/blob/devel/src/awx_plugins/credentials/hashivault.py#L451-L453

src/awx_plugins/credentials/hashivault.py
fincamd: This comment was removed to make the linter happy because that line was triggering the "comment line too long"

fincamd: Added back here 0b609500a3024170cd9c7ebfb0d6b1d2ae6aa424

PR #182, opened 2026-04-28
feat: support default credential chain for AWS Secrets Manager

yoonhyunwoo:
Make the access key and secret key optional to support default credentials (e.g., pod identity)
are used without exposing keys.

Also, ensure that if only one of the two is provided,
the operation fails naturally. Fallbacks to default behavior may confuse users.

## Summary by CodeRabbit

* **New Features**
  * AWS Secrets Manager credential plugin now treats access key and secret key as optional, allowing authentication via alternative mechanisms (e.g., IAM roles) when credentials are not provided.

### DIFF
--- a/src/awx_plugins/credentials/aws_secretsmanager.py
+++ b/src/awx_plugins/credentials/aws_secretsmanager.py
@@ -40,8 +40,6 @@ secrets_manager_inputs: _types.PluginInputs = {
         },
     ],
     'required': [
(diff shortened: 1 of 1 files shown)


yoonhyunwoo:
The test failure is not caused by this change

Related:
- #183

yoonhyunwoo:
Rebased onto latest devel

PR #188, opened 2026-06-10
Fix incorrectly labeled credential type

jessicamack:
The credential type for Ansible Automation Platform has a new field `request_timeout` that is labeled
“verify SSL” which is incorrect as according to the help text it's for timeouts, and has a default value of “galaxy” which causes an error as an int is expected. This PR corrects the field and adds a test to catch fields that are likely incorrectly typed. 

## Summary by CodeRabbit

* **New Features**
  * Added a request timeout option for controller credentials with a default of 10 seconds.

* **Bug Fixes**
  * Verify SSL option for controller credentials now consistently behaves as a boolean and no longer uses a preset default, requiring explicit selection or falling back to system behavior.


AlanCoding (approved):

src/awx_plugins/credentials/plugins.py
```diff
@@ -749,10 +749,10 @@
             },
             {
                 'id': 'request_timeout',
-                'label': gettext_noop('Verify SSL'),
-                'type': 'string',
+                'label': gettext_noop('Request Timeout'),
+                'type': 'int',
```
lennysh: `'int' is not one of ['string', 'boolean']`

src/awx_plugins/credentials/plugins.py
```diff
@@ -749,10 +749,10 @@
...
                 'secret': False,
-                'default': 'galaxy',
+                'default': 10,
```
lennysh: `10 is not a string`

src/awx_plugins/credentials/plugins.py
jessicamack: you're right. the error being on `str` type I defaulted to making sure it would be an `int`  for conversion but checking the docs again I see only string and boolean are an option https://docs.ansible.com/projects/awx/en/24.6.1/userguide/credential_types.html. fixing now

tests/test_timeout_field_types.py
lennysh, line 1: Type `string` and `boolean` are the ONLY two allowed types, so not sure this script is needed/going to help. Unless it just makes sure they are all of type `string` and not `boolean`, but that wouldn't of caught this issue as `string` is correct in this case.

jessicamack: Yes, we wanted to have a test that could catch if something like this happened in the future but it's probably better to leave it off for now and find a better solution after this is fixed.

AlanCoding (approved):

lennysh (approved):

PR #195, opened 2026-08-18
Gate Coveralls upload to the upstream repo to unblock fork CI

TheRealHaoLiu:
## Summary

Fork CI is currently red on every branch (including `devel`): pre-commit hooks all pass, but the job fails uploading MyPy coverage to Coveralls with `Unprocessable Entity / Couldn't find a repository matching this job`, since forks aren't registered on coveralls.io.

This makes the Coveralls upload's `fail-on-error` true only when the run is happening in the upstream repo (`pre-setup.outputs.upstream-repository-id == github.repository_id`), mirroring the Codecov gate the tests job already uses. `job-dependencies-context: ${{ toJSON(needs) }}` is forwarded into the lint job so the hook can read that `pre-setup` output.

In forks, the upload still attempts but its rejection no longer fails the job. Since `ansible/awx-plugins`'s repo id equals `UPSTREAM_REPOSITORY_ID`, the gate always evaluates to true here — this is a **behavioral no-op on this repo**: strict enforcement is preserved upstream, and only forks stop failing CI on the Coveralls step.

## Test plan

- [x] `pre-commit run` on the two changed files — all hooks pass except a pre-existing `actionlint`/pyflakes false positive on `ci-cd.yml:245` (an unrelated inline Python script) that is also present on unmodified `devel`.
- [x] Confirmed `ansible/awx-plugins`'s repository id matches `UPSTREAM_REPOSITORY_ID`, so the gate is unchanged (`true`) on this repo.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of code coverage upload failures, allowing debug and forked runs to continue while enforcing uploads for standard upstream runs.

* **Chores**
  * Enhanced continuous integration workflow context handling for linting jobs.
  * Removed the Coveralls status badge and related links from the project documentation.

### DIFF
--- a/.github/reusables/tox-dev/workflow/reusable-tox/hooks/post-tox-job/action.yml
+++ b/.github/reusables/tox-dev/workflow/reusable-tox/hooks/post-tox-job/action.yml
@@ -35,7 +35,16 @@ runs:
     uses: coverallsapp/github-action@v2
     with:
       debug: ${{ runner.debug == 1 && true || false }}
-      fail-on-error: ${{ runner.debug == 1 && false || true }}
+      # NOTE: Only require the upload to succeed in the upstream repo,
+      # NOTE: since forks are not registered on Coveralls.
+      fail-on-error: >-
+        ${{
+          runner.debug != 1
+          && fromJSON(
+            inputs.job-dependencies-context
+          ).pre-setup.outputs.upstream-repository-id
+          == github.repository_id
+        }}
       files: >-
         ${{
           fromJSON(
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -406,6 +406,8 @@ jobs:
         }}
       dists-artifact-name: >-
         ${{ needs.pre-setup.outputs.dists-artifact-name }}
+      job-dependencies-context: >-  # context for hooks
+        ${{ toJSON(needs) }}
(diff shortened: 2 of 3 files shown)


AlanCoding (approved):

AlanCoding:
https://github.com/ansible/awx-plugins/pull/196 would make this unnecessary. We will collect coverage with sonar instead. That will make this no longer necessary.

webknjaz:
> #196 would make this unnecessary. We will collect coverage with sonar instead. That will make this no longer necessary.

Not true. Sonar doesn't get type preciseness metrics. It was separated out into Coveralls because the overlay it in the same UI (with the test coverage) was confusing.

.github/reusables/tox-dev/workflow/reusable-tox/hooks/post-tox-job/action.yml
```diff
@@ -35,7 +35,16 @@ runs:
     uses: coverallsapp/github-action@v2
     with:
       debug: ${{ runner.debug == 1 && true || false }}
-      fail-on-error: ${{ runner.debug == 1 && false || true }}
+      # NOTE: Only require the upload to succeed in the upstream repo,
```
webknjaz, line 38: @TheRealHaoLiu were you getting this on `push` events or opening in-fork PRs?

